Issue #27, opened 2013-06-30
Use multiple threads to speed up baking


jonbullock:
Use multiple threads to speed up baking

PR #38, opened 2013-10-12
#27, multithread rendering for pages and posts.

s17t:


jonbullock:
First off I'd like to say thanks for taking the time to raise this PR in the first place.

Unfortunately I'm not going to merge this in, primarily due to the code base changing so much since it was opened. This is my fault for not dealing with it sooner, so I do apologise for that.

I do still want to improve the speed of JBake and your PR will serve as a great base for doing this at some point in the future.

Thanks again.

PR #72, opened 2014-01-05
added weld / CDI

RainerW:
Added JBoss weld (http://weld.cdi-spec.org/) for CDI support 

why:
- no static var problem in tests
- code gets testable
- future code can now use CDI to decouple stuff ( e.g. melix modularization code could use this )

did:
- added weld as CDI
- removed static code from ConfigUtil
- changed Tests so no static reset needed
- added mockito as dependency will be usedfull for nice test ( e.g. http://docs.mockito.googlecode.com/hg/org/mockito/InjectMocks.html )
- all changes from https://github.com/jonbullock/JBake/pull/71 ( becuase i'm lazy, not per se necessary )


jonbullock:
I'm going to close this off now. The code base has changed significantly since this PR was proposed so it's not going to be merged as it is anyway.

Thanks to everyone for participating in this discussion and providing valuable opinions, especially to @RainerW for the initial PR.

PR #78, opened 2014-01-17
Plugin bootstrap proof of concept

mwanji:
Not really meant to be pulled, just an example how a plugin system could be bootstrapped:
- A `plugin` directory is created next to `lib`. JARs can be dropped in here.
- `jbake` script adds `plugin` to classpath (`jbake.bat` not modified and only tested on Ubuntu, not Mac).
- Plugins implement the `JBakePlugin` interface and declare themselves as a service for that interface.
- In `Main`, a `ServiceLoader` picks up all implementations of `JBakePlugin` and calls `init` on each one.

### DIFF
--- a/src/main/assembly/assembly.xml
+++ b/src/main/assembly/assembly.xml
@@ -40,6 +40,13 @@
             </includes>
         </fileSet>
         <fileSet>
+            <directory>${project.build.directory}</directory>
+            <outputDirectory>plugin</outputDirectory>
+            <excludes>
+              <exclude>**/*</exclude>
+            </excludes>
+        </fileSet>
+        <fileSet>
             <directory>${basedir}</directory>
             <outputDirectory></outputDirectory>
             <includes>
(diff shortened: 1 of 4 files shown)


jonbullock:
Here's my original vision for plugins/extensions:

There will be a number of hook or extension points in the JBake API, each at various points along the processing pipeline. Such as content crawling & parsing (to include more content than just static files), pre-processing (allowing modification of content before format conversion say from Markdown to HTML), post-processing (allowing modification of the HTML to be rendered), rendering (allowing template engine related alterations), and asset processing (allowing minification of CSS & JS or conversion from LESS to CSS). Each of these points will have their own plugin interface to allow multiple types of plugins, so there can be multiple plugins for each extension point. What each plugin can do depends on the interface for that extension point in the processing pipeline, so at the content crawling and parsing stage the plugin can interact with the content store. At the asset processing stage the plugin will be passed each asset file and it's up to the plugin to decide whether to perform additional processing on the file or not.

I realise this is a proof of concept and slightly different to what I've just outlined however I can see having an init method potentially beneficial. I also like the idea of having a separate plugins folder.

Anyway I just wanted to get my vision out into the public so it can be discussed, changed, fixed or updated. All comments welcome!

I'll also knock up some diagram and post this to the dev mailing list too.

PR #104, opened 2014-03-26
Extensionless URLs

mwanji:
New property `uri.noExtension` allows the user to specify a root folder for which "extensionless" URLs will be created. In practice, the file name is used as a folder and the actual file is generated as index.html inside that folder. `uri.noExtension` defaults to `false`, which does nothing.

For example:

If `uri.noExtension` is set to `/blog`, then after baking, the file `/blog/2014/03/26/post.md` would become `/output/blog/2014/03/26/post/index.html` and the URL would be `<DOMAIN>/blog/2014/03/26/post`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
 		<commons.configuration.version>1.9</commons.configuration.version>
 		<args4j.version>2.0.23</args4j.version>
 		<freemarker.version>2.3.19</freemarker.version>
-		<junit.version>4.10</junit.version>
+		<junit.version>4.11</junit.version>
 		<pegdown.version>1.4.2</pegdown.version>
 		<jetty.version>8.1.12.v20130726</jetty.version>
         <orientdb.version>1.6.2</orientdb.version>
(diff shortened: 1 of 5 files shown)


jonbullock:
I like this idea... not sure if limiting this to just a single folder is the right way to go though? Is this something that should be defined for a whole site/project?

puredanger:
I have the same desired goal as this (urls with no extension), but would actually prefer to just create a file with no extension at all. This works fine on s3 (for example) if you set the content-type property for the file.

jonbullock:
Hi Alex, try defining `output.extension` in your jbake.properties without a value to override the default value of `.html` which should achieve what you want. I could be wrong as I've not had time to try this myself.

puredanger:
I have tried that - it doesn't work.

jonbullock:
Ahh I see what the problem is, I've just committed a fix to master branch, it'll be included in v2.5.0 release. The built in pages will still get an extension but they can be overridden by using additional options.

puredanger:
Is there an expected timeframe for the 2.5.0 release?

jonbullock:
I'm working on it at the moment and I'm hoping to have it out by year end at the latest. This is the last pull request I need to deal with then it's clean up work, as the master branch has moved on significantly since some of the pull requests were created.

PR #105, opened 2014-03-27
Create files and metadata from the command line

mwanji:
I always have trouble remembering the correct metadata when creating a new file and how many tildes are in the separator.

`jbake -c my-post.md` creates a file called my-post.md in the current directory. It then gathers metadata, similar to `npm init`.

This is not quite complete yet (e.g., no asciidoc support). If @jonbullock thinks there's a good chance of integrating this feature, I'll add asciidoc, some more unit tests and whatever else is necessary.


jonbullock:
@mwanji are you still working on this?

PR #106, opened 2014-03-28
Localization POC

bashnesnos:
Hello there,

While I've been playing around with JBake, I just thought may be a localization feature could be useful?

The idea is: localized pages during the build are being put to a separate folder. Since the pages are static it is the cheapest and straight-forward option. Switching between the localized/original is to be operated by URL manipulation in javascript. E.g. the following transformation will occur:

```
src\jbake\content\blog\2014\jbake_locale.md -> \blog\2014\jbake_locale.html (original)
src\jbake\content\blog\2014\en_jbake_locale.md -> \en\blog\2014\jbake_locale.html (localized)
```

Overall convention is the following:
- Localized post file name should differ from original only by _localeId__ prefix (in my case **en_**)
- Add the localized templates into jbake.properties
- Localized template name should start from _localeId__ (in my case **en_**)
- Keep the _localeId__ in mind when populating links and includes inside of the templates

If none are supplied - no localization, falls back to the usual transformation.

If the idea is all right, I'll extend the current POC to all other template engines etc.

P.S. working blog [sources](https://github.com/bashnesnos/blog) and live [example](http://bashnesnos.github.io/blog/en/)

### DIFF
--- a/src/main/java/org/jbake/app/Crawler.java
+++ b/src/main/java/org/jbake/app/Crawler.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.Set;
 
 import static java.io.File.separator;
+import org.jbake.model.LocaleTypes;
 
 /**
  * Crawls a file system looking for content.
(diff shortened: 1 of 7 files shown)


uli-heller:
@bashnesnos: I like the idea, but unfortunately. I cannot use it at the moment because:

1. The description seems to be a little bit incorrect from my understanding
    * '\' --> '/' (urls don't use backslashes)
    * 'en/blog/...' -> 'blog/en/...' (I've been looking at the [live example](http://bashnesnos.github.io/blog/en/) for this)
2. There seem to be merge conflicts ;)
3. I'm using GSP templates :(

I'm doing by blog in German at the moment. For selected posts, I'm using English. I guess this is not a good use-case for your approach anyway, isn't it?

PR #113, opened 2014-05-24
Adds publisher option to jbake

lordofthejars:
With this pull request jbake users can create their own publishers modules so after a site is baked can be published automatically.

Currently a Github publisher has been implemented but users can implement their own by using the jbake spi.

Also JDK has been updated to Java 7 to be able to use Java IO features developed on that version of java.

### DIFF
--- a/jbake.properties
+++ b/jbake.properties
@@ -1,3 +1,0 @@
-site.host=http://jbake.org
-render.tags=false
-render.sitemap=true


msgilligan:
Personally, I'd rather JBake *not* add publishing modules. I'm using the Gradle Github Publisher and it seems to me that it makes more sense to use your build tool of choice (e.g. Gradle, Maven, etc) to do the  publishing step. I think there will be a greater selection of higher-quality choices that way and JBake can stay focused on core functionality.

aalmiray:
I agree with @msgilligan. I too use the Gradle Github Publisher. Adding publishing capabilities to jbake might make it overextend its responsibilities. The advantage of having it inside jbake is that any build tool could extend it with custom publishers. OTOH jbake is a site rendering tool (so far).

I feel it for the maven users because the gradle plugin ecosystem is much more fun :smirk:

jonbullock:
Thanks for the comments guys. I'm torn about adding this feature, which is why it's been pushed back to v3.0.0, for users who use build systems it does overlap existing functionality available. However, for users who prefer the CLI it would be beneficial. What if this was provided by a plugin/extension instead? Would allow JBake to stay focused on core functionality and this would not then become a consideration in any future development...

aalmiray:
Pluggable publishing extensions sound like a good compromise to me :smirk:

PR #146, opened 2014-09-30
created two new engines to simplify work with html content

Vad1mo:
this are some ideas I was working on.  This is not production ready just to have some ideas.

Scalate or the new GroovyTemplateEngine could be used not only on the template side but also on the engine side or event both.

One cool thing is that you can mix Scalate with markdown, would be nice to have it also for GroovyTemplateEngine

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,7 +113,7 @@
 						</goals>
 						<configuration>
 							<checkoutDirectory>${project.build.directory}/example-project-freemarker</checkoutDirectory>
-							<connectionUrl>scm:git:git://github.com/jbake-org/jbake-example-project-freemarker.git</connectionUrl>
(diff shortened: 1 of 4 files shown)


jonbullock:
Just been reviewing this pull request, have you got sample content files using Scalate & Groovy MarkupTemplateEngine for me to see?

I had originally wanted to use Groovy MarkupTemplateEngine as a template format but hadn't considered it as a content format.

I'm going to bump this to v3.x for the time being since it's not production ready yet.

danhyun:
I'm looking forward to having the `MarkupTemplateEngine` as well. Looks like @ancho is working on this https://github.com/ancho/jbake/blob/feature/groovy-MarkupTemplateEngine/src/main/java/org/jbake/template/GroovyTemplateEngine.java#L59

jonbullock:
@Vad1mo Have you done any more on this? If not I'll close this for the time being.

PR #162, opened 2014-12-17
Fix copy of assets

HuguesH:
Hello, thanks very for the only one static site generator in JAVA.
During my test of Jbake, I ve change the assets directory by configuration, but the copy of assets is broken.
I have find a Fix, thank's in advance to look this problem if my code is not correct.


jonbullock:
Thanks for raising this issue, I've identified the root cause and fixed it.

PR #163, opened 2014-12-19
Thymeleaf templates are plain HTML files, they should be named .html

christianbauer:
This is important for example when live editing (a plugin) in IntelliJ. This doesn't work with .thyme extension and the IntelliJ built-in debug webserver.


jonbullock:
Thanks for the contribution.

PR #169, opened 2015-01-22
ContentStore

jasondlee:
This branch moves all DB-related code behind a facade, ContentStore. This work lays the groundwork for implementing the pagination feature so many have requested.

### DIFF
--- a/src/main/java/org/jbake/app/ContentStore.java
+++ b/src/main/java/org/jbake/app/ContentStore.java
@@ -23,19 +23,15 @@
  */
 package org.jbake.app;
 
-import com.orientechnologies.orient.core.command.OCommandRequest;
 import com.orientechnologies.orient.core.db.ODatabaseRecordThreadLocal;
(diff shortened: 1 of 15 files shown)


src/main/java/org/jbake/app/ContentStore.java
```diff
@@ -0,0 +1,167 @@
...
+        return query("select * from "+docType+" where status='published' order by date desc");
+    }
+
+    public List<ODocument> getAllContent(String docType) {
+        return query("select * from "+docType+" order by date desc");
+    }
+
+    public List<ODocument> getAllTags() {
+        return query("select tags from post where status='published'");
+    }
```
lefou: That method name should reflect the fact that the tags are extracted from published posts.

jasondlee: Changed and pushed.

jonbullock:
Thanks for the contribution.

PR #171, opened 2015-02-03
Markdown.extensions: Added STRIKETHROUGH, enhanced configuration of markdown.extensions: ALL,-HARDWRAPS

uli-heller:
This change enhances the configuration of the markdown.extensions.
So far, you could just specify all the active extensions. With this
change, you can specify unwanted extensions as well.

For example:

    markdown.extensions=ALL,-HARDWRAPS

activates all extensions with exception of the HARDWRAP extension


jonbullock:
Thanks for the contribution.

PR #175, opened 2015-02-11
Simplify and aggregate rendering testing

jasondlee:
In getting ready to implement pagination, I felt I needed to test the Freemarker template engine, as that's what I intended to use (so I want to make sure it works ;). As I was working on it, I realized all of the tests for the template engines were the same, so I refactored the code so that they share as much as possible. There are some minor differences in the resulting HTML (that might warrant some investigation) thus the code in the ctors, but I didn't see that as a show stopper for integrating the current state of the refactoring.

Create an abstract base test
Extend the base class for each template engine


jonbullock:
Thanks again Jason for the contribution. I should have some time next week to review this.

PR #193, opened 2015-05-07
Enable asset copy to ignore hidden resources

pitpalme:
During normal usage I noticed an automatically created file ".DS_Store" was copied into destination folder. Of course it's not jbake's fault, but on the other hand you only have to open the assets folder using "Finder" to risk this file being created without one noticing.

So I thought it might be nice to have an option to simply ignore those automatically created (and usually hidden) files.

### DIFF
--- a/src/main/java/org/jbake/app/Asset.java
+++ b/src/main/java/org/jbake/app/Asset.java
@@ -1,6 +1,7 @@
 package org.jbake.app;
 
 import java.io.File;
+import java.io.FileFilter;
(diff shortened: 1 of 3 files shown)


jonbullock:
Thanks for the contribution Peter.

pitpalme:
Pardon for silently extending the PR. It wasn't my intention when I clicked around in GitHub GUI, but as it's harmless and kind of related I didn't try to delete enhancement until original PR is dealt with ...

Issue #194, opened 2015-05-07
API change for Oven constructor


pitpalme:
Hello,

I've noticed there's an API change after 2.3.2, introduced in commit a4458cb83b89d30d7f64ad00cb6caedff467e59d.

org.jbake.app.Oven's constructor was enhanced by on additional parameter: 'config'.

This is problematic when using https://github.com/ingenieux/jbake-maven-plugin/ in it's current version 0.0.9.

Is there any specific reason there's no delegating constructor preserving the old API, which if present would enable usage of post jbake-core-2.3.2 in connection with jbake-maven-plugin-0.0.9?

I'd volunteer to submit an appropriate commit ...

Regards,
Peter

jonbullock:
Hi Peter,

Thanks for picking this one up, the original commit that made this change was: https://github.com/jbake-org/jbake/commit/cef489d5582ea38275df3e1f320211a52c6d2432

This API change will also cause a similar issue for the Gradle plugin too.

Moving forward I'd like to avoid having a delegating constructor if we can address the underlying need for it.

@lefou is there a specific reason the config shouldn't be initialised in the Oven constructor?

lefou:
The configuration is used in multiple places. Now, it is loaded only once in the `Main.run()` method. By injecting it into various places it is guaranteed that all components use the same configuration. Also is would be easy to change the way the configuration is loaded. In fact, it doesn't need to be loaded from a file at all. This of course supports our aim to use JBake as a library.

Why is the Gradle plugin using the Oven class at all? Maybe, it's time to start to think about a proper JBake API?

jonbullock:
Thanks @lefou I may be wrong about the Gradle plugin, I was looking at this line: https://github.com/jbake-org/jbake-gradle-plugin/blob/master/src/main/groovy/me/champeau/gradle/JBakeTask.groovy#L70

The Main class and CLI are wrappers around the core API of which Oven is the main entry point and it has crossed my mind to improve the core API to make it more friendly (i.e. removing the need for Oven.setupPaths() to be called) for those using JBake just as a library (i.e. the Maven/Gradle plugins among others). My preference would be for the config to be loaded from within Oven so it's transparent in terms of the API and only loaded once, however the Main class & CLI currently require certain info from the default.properties which prevents this from happening right now. Perhaps this should be the goal? To remove the need of the Main class & CLI having access to the config and allowing Oven to take care of it.

What do you both think?

jonbullock:
[...]

PR #198, opened 2015-05-09
API restore

pitpalme:
As discussed in #194 and requested there, here's the PR


jonbullock:
Thanks, could you remove the asset copy commits and unit test commits from this PR?

jonbullock:
Actually on second thought, don't worry about it.

PR #212, opened 2015-06-07
Fixed link to groovy - http://groovy-lang.org is the new place

uli-heller:


jonbullock:
Thanks for catching this.

PR #218, opened 2015-06-25
Detailed errors with stack trace of the original problem

haumacher:
If jbake crashes with an exception, no detailed stack traces are reported. If  exceptions are thrown e.g. during rendering, these exceptions are caught and their messages are recorded. Afterwards, a fresh exception is thrown with all aggregated messages concatenated - so far so good. The problem is that the new exception has no linked cause making debugging a nightmare. It would be better to not record the messages but the exceptions themselves, build the new exception message from their messages, but at least link the first cause as cause of the new exception.

Besides that, it is good practice to link wrapped exceptions to their cause. Whenever one has the pattern 

```java
} catch (AException ex) {
   throw new BException("...");
}
```
this should be re-written as 
```java
} catch (AException ex) {
   throw new BException("...", ex);
}
```

### DIFF
--- a/src/main/java/org/jbake/app/Asset.java
+++ b/src/main/java/org/jbake/app/Asset.java
@@ -26,7 +26,7 @@ public class Asset {
     private File source;
 	private File destination;
 	private CompositeConfiguration config;
-	private final List<String> errors = new LinkedList<String>();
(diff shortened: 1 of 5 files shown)


haumacher:
Wrong branch (with unrelated other pull-request as base branch).

PR #219, opened 2015-06-25
Detailed errors with stack trace of the original problem

haumacher:
If jbake crashes with an exception, no detailed stack traces are reported. If  exceptions are thrown e.g. during rendering, these exceptions are caught and their messages are recorded. Afterwards, a fresh exception is thrown with all aggregated messages concatenated - so far so good. The problem is that the new exception has no linked cause making debugging a nightmare. It would be better to not record the messages but the exceptions themselves, build the new exception message from their messages, but at least link the first cause as cause of the new exception.

Besides that, it is good practice to link wrapped exceptions to their cause. Whenever one has the pattern 

```java
} catch (AException ex) {
   throw new BException("...");
}
```
this should be re-written as 
```java
} catch (AException ex) {
   throw new BException("...", ex);
}
```

### DIFF
--- a/src/main/java/org/jbake/app/Asset.java
+++ b/src/main/java/org/jbake/app/Asset.java
@@ -26,7 +26,7 @@ public class Asset {
     private File source;
 	private File destination;
 	private CompositeConfiguration config;
-	private final List<String> errors = new LinkedList<String>();
+	private final List<Throwable> errors = new LinkedList<Throwable>();
 	private final boolean ignoreHidden;
 
 	/**
@@ -70,7 +70,7 @@ public class Asset {
 						sb.append("failed!");
 						LOGGER.error(sb.toString(), e);
 						e.printStackTrace();
-						errors.add(e.getMessage());
+						errors.add(e);
 					}
 				}
 
@@ -81,8 +81,8 @@ public class Asset {
 		}
 	}
 
-	public List<String> getErrors() {
-		return new ArrayList<String>(errors);
+	public List<Throwable> getErrors() {
+		return new ArrayList<Throwable>(errors);
 	}
 
 }
--- a/src/main/java/org/jbake/app/Oven.java
+++ b/src/main/java/org/jbake/app/Oven.java
@@ -40,7 +40,7 @@ public class Oven {
 	private File contentsPath;
 	private File assetsPath;
 	private boolean isClearCache;
-	private List<String> errors = new LinkedList<String>();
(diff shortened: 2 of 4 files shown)


jonbullock:
Sounds like a good improvement... have you got an example you've tested this with so I can see what it looks like in practice?

haumacher:
When something goes wrong during development (e.g. a NullPointerException), you get the following output with the original version:

	JBake failed with 35 errors:
	1. Failed to render file. Cause: null
	...
	35. Failed to render tags. Cause(s):
	null
	null
	null
	null
	
	org.jbake.app.JBakeException: JBake failed with 35 errors:
	1. Failed to render file. Cause: null
	...
	35. Failed to render tags. Cause(s):
	null
	null
	null
	null
	
		at org.jbake.launcher.Main.bake(Main.java:62)
		at org.jbake.launcher.Main.run(Main.java:86)
		at org.jbake.launcher.Main.main(Main.java:36)

This gives absolutely no clue about the location of the error. With the change included the output looks like the following:

	
	org.jbake.app.JBakeException: JBake failed with 35 errors:
	1. Failed to render file. Cause: null
	...
	35. Failed to render tags. Cause(s):
	null
	...
	null
	
		at org.jbake.launcher.Main.bake(Main.java:62)
		at org.jbake.launcher.Main.run(Main.java:86)
		at org.jbake.launcher.Main.main(Main.java:36)
	Caused by: java.lang.Exception: Failed to render file. Cause: null
		at org.jbake.app.Renderer.render(Renderer.java:96)
		at org.jbake.app.Oven.bakeIncremental(Oven.java:193)
		at org.jbake.app.Oven.bake(Oven.java:164)
		at org.jbake.launcher.Main.bake(Main.java:50)
		... 2 more
	Caused by: java.lang.NullPointerException
		at org.jbake.template.ThymeleafTemplateEngine.renderDocument(ThymeleafTemplateEngine.java:87)
		at org.jbake.template.DelegatingTemplateEngine.renderDocument(DelegatingTemplateEngine.java:65)
		at org.jbake.app.Renderer.renderDocument(Renderer.java:111)
		at org.jbake.app.Renderer.render(Renderer.java:89)
		... 5 more

Which easily shows, that something is wrong invoking the ThymeleafTemplateEngine.

jonbullock:
Thanks for taking the time to highlight the benefits. I'll update the PR when I know which release this will be included in.

PR #220, opened 2015-06-25
Superfast baking with Thymeleaf

haumacher:
Currently the Thymeleaf  template engine is re-initialized for each and every generated page. Initializing the template engine is a costly operation, because this drops the resolved templates. When initializing the template engine for each page, all used templates must be resolved for each rendering operation. This greatly reduces rendering performance. Instead, the template engine should only be only re-initialized, if the template mode changes.

With this change, baking time for my relatively small recipe dramatically drops from over 13s to around 7s. But I expect an even more dramatic performance boost for larger scenarios.


jonbullock:
Performance improvements are always welcome :+1: 

Thanks for the contribution.

PR #229, opened 2015-09-11
Support absolute paths for asset, content and template folders

acanda:
These changes allow both relative and absolute paths for the configuration properties `asset.folder`, `content.folder` and `template.folder`.

This makes the configuration for the jbake-maven-plugin much more readable as you can use `<content.folder>${project.build.directory}/jbake/content</content.folder>` instead of something like `<content.folder>../../../target/jbake/content</content.folder>`.


jonbullock:
Thanks for the contribution.

Issue #180, opened 2015-02-19
Make the parsing metadata more flexible(intelligent)


gilbertoca:
Currently the following header is not recognizable by jbake [1]:
```

title= Removendo alertas antigos da página principal do DBCONSOLE
date= 2007-10-08 14=22=17
type= post
status= publish
tags=
,Concurso Público
,Database
,dbconsole
,oracle
```
It complains about the following items:

1. the first line is blank
2. each field has white space after equal sign
3. the field tags is multi-line 

It would be good if it was more flexible. Maybe performing a trim after the equal sign and checking only the final value for correctness -in the multi-line case, concatenate all line.

[1] https://github.com/gilbertoca/gilbertoca.com/blob/d4d5e67b2b0db4dc8a7cf158bb7c6f0612e9ee88/src/jbake/content/2007/10/2007-10-08-removendo-alertas-antigos-da-pagina-principal-do-dbconsole.md

jonbullock:
I agree the header code should be more tolerant of blank lines and spaces and this will be included in v2.5.0, but I don't agree with the multiline option support. The header is meant to be simple and concise.

PR #241, opened 2015-11-06
Make the parsing metadata more flexible

atao60:
This request answers to the three enhancements suggested by [Make the parsing metadata more flexible(intelligent) #180](https://github.com/jbake-org/jbake/issues/180):
* Allow blank lines in header of content files
* Allow spaces around equal of a header entry
* Allow multi-lines value for header entry value  

About this last point: any line starting with 2 spaces will be concatenated with the previous line.

### DIFF
--- a/src/main/java/org/jbake/app/Parser.java
+++ b/src/main/java/org/jbake/app/Parser.java
@@ -1,5 +1,7 @@
 package org.jbake.app;
 
+import static org.jbake.app.Parser.ContentBasicTags.*;
+
 import org.apache.commons.configuration.CompositeConfiguration;
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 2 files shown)


jonbullock:
Thanks for the contribution.

jonbullock:
I've been reviewing this PR and I've come to conclusion that I won't be taking the change in. The reason being I don't want to complicate the structure of the metadata header, it's simple as is and while this fixes #180 I don't agree with part of the original change to allow multi line options. The structure of the header is clearly set out, one line = one option. Having said that, the other two suggestions I see as making the parsing code more tolerant which can be addressed with a couple of small code changes.

Thanks again for taking the time to contribute and apologies for the time it's taken in reviewing this PR.

PR #243, opened 2015-11-13
Add a smart excerpt/summary generator.

atao60:
Add a smart generator of excerpts for index and feed pages.

If no summary is provided by a post header, jbake will create excerpts from the post body.

The excerpts are ready to be used in index and feed pages. 
No more hard coded adaptations are required in templates.

The generator is quite versatile:
* length and counter type (word, character or Unicode code point),
* ellipsis,
* readmore pattern.

Warning. Some refactoring was made in the process, mainly the creation
of:
* a Content class to wrap the Map<String, Object> used by the Parser,
* a PostParsingProcessor class to use the post uri inside the summary/excerpt,
* a ContentStatus enumeration to gatherthe post/pages status  together, 
* a ContentTag enumeration to gather the labels used with Content.

It will interfere with [Preliminary refactoring #135](https://github.com/jbake-org/jbake/pull/135).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,13 @@
         <thymeleaf.version>2.1.3.RELEASE</thymeleaf.version>
 		<thymeleaf.extras.version>2.1.1.RELEASE</thymeleaf.extras.version>
         <json-simple.version>1.1.1</json-simple.version>
+        <tika.version>1.9</tika.version><!-- Version 1.9 of Tika is the last one working with Java 6 -->
+        <tagsoup.version>1.2.1</tagsoup.version>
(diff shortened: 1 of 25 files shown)


jonbullock:
Wow that is quite a pull request!

I haven't had chance yet to review all of the code but the first question that came to mind was how can you configure the generator to use one of the various strategies you list?

Also, seeing PostParsingProcessor in your description, this is one of the hook points I want to provide for the plugin/extension system. 

I'll need to think a bit more about when and how this is integrated.

atao60:
Through jbake.properties. See e.g. [here](https://github.com/atao60/pop-tech/blob/master/src/jbake/jbake.properties), at the end of the file.

PR #247, opened 2015-11-20
fixed broken template engine unit tests

ancho:
The FreemarkerTemplateEngineRenderingTest failed. Fixed template files and refactored the AbstractTemplateEngineRenderingTest a bit.

### DIFF
--- a/src/test/java/org/jbake/app/template/AbstractTemplateEngineRenderingTest.java
+++ b/src/test/java/org/jbake/app/template/AbstractTemplateEngineRenderingTest.java
@@ -27,10 +27,8 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.util.HashMap;
-import java.util.Iterator;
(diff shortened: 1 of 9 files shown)


jonbullock:
Thanks @ancho

jonbullock:
Long shot I know, but I don't suppose you remember what the test failed on do you?

ancho:
We had common (duplicated) expectations on the rendered output for all template engines. I moved them to the AbstractTemplateEngineRenderingTest.

The Freemarker templates used different html elements which caused the FreemarkerTemplateEngineRenderingTest to fail.

But thinking it through it is maybe not the best idea to define a common set of expectations on the rendering output as it dictates the structure of all templates.

It's maybe better to change the expectations of the concrete Test.

jonbullock:
The tests are dependant on the templates in the src/test/resources regardless and at the moment these tests are more integration tests to ensure the template engine code is working as expected. I'm happy to alter the structure when there is a need to test a different expectation of one of the template engines.

PR #248, opened 2015-11-20
Fixed loading of RenderingTool Providers

ancho:
Renamed provider-configuration file according to https://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html and wrote a little test.


jonbullock:
Thanks @ancho hadn't noticed this when I raised and fixed #254, but the test class is very useful!

Issue #231, opened 2015-10-13
Make metadata header separator configurable "~~~~~~" or support standards such as Jekyll front matter


sengopal:
Make metadata header separator configurable "~~~~~~" or support standards such as Jekyll front matter

Reason: I am using Atom.io and markdown-preview-plus for drafting posts. The separator "~~~~~~" seems to render the markdown as incorrect and the preview fails.

<img width="1439" alt="screen shot 2015-10-12 at 4 59 57 pm" src="https://cloud.githubusercontent.com/assets/787381/10442557/54e4b164-7108-11e5-92d6-211e13220916.png">

Raised it on markdown-preview-plus as well - https://github.com/Galadirith/markdown-preview-plus/issues/134

jangalinski:
the ~ is also hard to type on a macbook (ger) so +1 for making this more standard

sengopal:
Jekyll head metadata is with ---------. Even replacing or accepting ---------- will also help.

jonbullock:
Adding support for --------- should be straight forward to do. Have you tried a file with the JBake separator replaced with ---------  to check it works? 

As I thought Jekyll's YAML front matter starts with --- and ends with --- ?

sengopal:
If i try with --------- , the file does not get picked up for conversion. Do you think Jekyll's format will work, i can test it out.

jonbullock:
Yeah try the Jekyll format to confirm, adding support for the full front matter tags is a bit more work compared to changing the separator.

PR #249, opened 2015-12-05
Minor refactoring

thehereward:
A small amount of refactoring:
 - corrected a spelling
 - introduced a constant to replace a magic string
 - simplified some boolean logic
 - moved some functionality into small methods.

This reveals a couple of small issues which I may get time to address in other pull requests:
 - The 'isJson' method should probably be changed to be a little more robust in checking whether the contents is json
 - The header processing logic looks like it could be moved into a class of its own


jonbullock:
Thanks @thehereward - linking this with #231

PR #256, opened 2016-01-07
(GroovyTemplates) Use English Locale in the date formatter

jmini:
The [function `Date#format(java.lang.String)`](http://docs.groovy-lang.org/latest/html/groovy-jdk/java/util/Date.html#format(java.lang.String)) used in the GroovyTemplates uses the system Locale to format the date. When the Date format uses `MMMM` or `MMM` the name (resp. the short name) of the month is used.

This make the JBake build platform dependent (for example: the locale configuration on my machine is German and it is English on my CI Server)

In my opinion, for those cases, the complete form of the formatter should be used:

    new java.text.SimpleDateFormat(<Pattern>, Locale.ENGLISH).format(<Date>)

This pull request makes the corresponding changes. If you prefer a different solution, I would be happy to implement it. Please provide some pointers in order for me to start.

### DIFF
--- a/src/test/resources/groovyTemplates/archive.gsp
+++ b/src/test/resources/groovyTemplates/archive.gsp
@@ -14,15 +14,15 @@
             <%def last_month=null;%>
             <%posts.each {post ->%>
 				<%if (last_month) {%>
-					<%if (post.date.format("MMMM yyyy") != last_month) {%>
-						<h3>${post.date.format("MMMM yyyy")}</h3>
+					<%if (new java.text.SimpleDateFormat("MMMM yyyy", Locale.ENGLISH).format(post.date) != last_month) {%>
+						<h3>${new java.text.SimpleDateFormat("MMMM yyyy", Locale.ENGLISH).format(post.date)}</h3>
 					<%}%>
 				<% } else { %>
-					<h3>${post.date.format("MMMM yyyy")}</h3>
+					<h3>${new java.text.SimpleDateFormat("MMMM yyyy", Locale.ENGLISH).format(post.date)}</h3>
 				<% }%>
 				
 				<h4>${post.date.format("dd MMMM")} - <a href="${post.uri}">${post.title}</a></h4>
-				<%last_month = post.date.format("MMMM yyyy")%>
+				<%last_month = new java.text.SimpleDateFormat("MMMM yyyy", Locale.ENGLISH).format(post.date)%>
 			<%}%>
 		</div>
 	</div>
--- a/src/test/resources/groovyTemplates/feed.gsp
+++ b/src/test/resources/groovyTemplates/feed.gsp
@@ -6,8 +6,8 @@
     <atom:link href="http://jonathanbullock.com/feed.xml" rel="self" type="application/rss+xml" />
     <description>My corner of the Internet</description>
     <language>en-gb</language>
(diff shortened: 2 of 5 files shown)


mikolak-net:
+1 for this. Just had to work around a problem where an RSS aggregator was malfunctioning due to non-English-localized dates present on my feed.

Of course, a better solution would be to have the date formatting local configurable, but this PR is a good first step.

jonbullock:
Thanks @jmini - just to double check your goal of this PR was to fix the tests? Not the example Groovy templates for creating a new project/site?

I'd be happy if somebody wants to make the Locale used for Date formatting configurable via jbake.properties like the Date pattern is currently.

jmini:
I wanted to update the Groovy templates used when a new project is created.
=> from your comment I understand that I am not at the correct location.

`jbake.properties` also seems a good Idea, but I think that this is for the templates that are displayed to the user (archive, index, post, tags). The feed template always need to use english (see @mikolak-net comment).

jonbullock:
Ahh in that case this is the repo you want: https://github.com/jbake-org/jbake-example-project-groovy

jmini:
I have opened a new pull request in the other repository: [jbake-example-project-groovy/#2](https://github.com/jbake-org/jbake-example-project-groovy/pull/2)

jonbullock:
Thanks I'll still take this change in to ensure the tests act the same way regardless of the machine running them.

ancho:
Hmm. Maybe we should configure the locale in ```default.properties``` like

```
locale=en_GB
```

We could then set the default locale in ```Oven``` or in ```Main``` after loading the configuration with
```Locale.setDefault(new Locale( config.getString("locale"));```.  What do you think?

mikolak-net:
@ancho : there should be a separate setting for the global locale, and another one for machine-readable stuff such as RSS.

ancho:
@mikolak-net like rendering all the pages with french locale settings and rss with english locale settings?

maybe optional locale settings per file? like ```feed.file=feed.xml``` and ```feed.locale=en_US```?

mikolak-net:
@ancho well, regarding the feed, after a quick read of the [RSS Spec](http://www.rssboard.org/rss-specification), I no longer think that a user-configurable option is the way to go in this case. It appears that valid RSS always requires non-localized dates in an [RFC822](http://www.faqs.org/rfcs/rfc822.html)-compatible format.

ancho:
Well, thank you. 

But that is not the only syndication format out there. The atom syndication format requires [rfc3339](http://tools.ietf.org/html/rfc3339) time and date formats as described in [rfc4278 date constucts](http://tools.ietf.org/html/rfc4287.html#section-3.3) 

I think it's best to make it configurable by type. Not just the locale but the dateformat as well.

mikolak-net:
@ancho you're right. And sorry, I completely neglected to check out the Atom spec. Yeah, the solution you proposed doesn't sound like YAGNI anymore, so :+1: from me.

ancho:
@mikolak-net Nothing to feel sorry about. Thanks for the discussion and your effort. You pointed me and my thoughts in the right direction. I call that teamwork :)
I'll open an issue with a feature request derived from our discussion (maybe tomorrow).

mikolak-net:
Glad to hear that :).

Issue #259, opened 2016-01-17
org.jbake.app.template.FreemarkerTemplateEngineRenderingTest.renderPost fails when non-English locale is used


iryndin:
Unit test `org.jbake.app.template.FreemarkerTemplateEngineRenderingTest.renderPost` fails when non-English locale is used. This is because FreeMarker uses local locale to render date strings, while unit test have hardcoded date string in English, which leads to failure. 

How to fix: do not hardcode date strings in unit tests, instead use generated date strings (e.g. with `SimpleDateFormat`).

iryndin:
here is pull request: https://github.com/jbake-org/jbake/pull/260

jonbullock:
Thanks for raising this, I'm thinking that the locale used by template engines should be explicitly set, there is already code in place to allow the locale to be set for Thymeleaf. We can always fall back to the default if one isn't set explicitly.

PR #260, opened 2016-01-17
Fix issue #259: do not use hardcoded date string, instead generate it dynamically

iryndin:


jonbullock:
Thanks for taking the time to raise this PR, would you be interested in adding support for a locale to be set in the template engine via configuration like what's in place for Thymeleaf?

jonbullock:
I believe this issue has been fixed thanks to this commit: https://github.com/jbake-org/jbake/commit/516d6c675f9ae4ec64f4bbb1c88b36f2737fe4ad

PR #286, opened 2016-06-11
register model extractors for custom types 

### LINKED ISSUE #283: custom types are broken
While integrating the groovy-mte example project (see #282) I stumbled upon a problem with custom templates/types.

They are not recognized anymore and are not accessible within the templates.

The new document type is added to the ```DocumentTypes```. It seems to be connected with the ModelExtractors mechanism. There is no registered model extractor for the new custom type.

Maybe the _[should never happen](https://github.com/jbake-org/jbake/blob/master/src/main/java/org/jbake/template/GroovyMarkupTemplateEngine.java#L78)_ part happened ;)

Needs more investigation.

ancho:
This is an attempt to fix #283 
It register model extractors for custom types during DelegatingTemplateEngine construction.
I couldn't resist to move a few things around and refactor a bit here and there.

The `renderCustomTypePaper` Test in `GroovyMarkupTemplateEngineRenderingTest` could be moved to the  `AbstractTemplateEngineRenderingTest` to test the behaviour of all template engines.


src/main/java/org/jbake/template/DelegatingTemplateEngine.java
```diff
@@ -30,8 +32,21 @@
     public DelegatingTemplateEngine(final CompositeConfiguration config, final ContentStore db, final File destination, final File templatesPath) {
         super(config, db, destination, templatesPath);
         this.renderers = new TemplateEngines(config, db, destination, templatesPath);
+        this.registerExtractorsForCustomTypes();
```
jonbullock: Was thinking about this.... could this call happen when other types are registered by Oven in DocumentTypes or was there a specific reason this occurs here?

jonbullock:
I noticed this adds `published_<type>s` to the data model as well, I'll have to remember to update the docs to include details about this as well.

src/main/java/org/jbake/template/DelegatingTemplateEngine.java
ancho: There was no special reason to add this in the DelegatingTemplateEngine.
It was just convenient. 
The Renderer is the class which instantiates the DelegatingTemplateEngine and is the first AbstractTemplateEngine which instantiates a ModelExtractors class filling up the shared Datastructure holding the references to the different extractors. At this time no new DocumentTypes get registered. That's it.
But you are right, maybe it is better to register the extractors after the DocumentTypes are collected, while warming up the Oven.
It would be more consistent to the Oven analogy.

* heat up
* bake
* cool down

ancho:
I rebased my branch to the latest master to make the merge easier.

jonbullock:
Thanks @ancho - much appreciated.

Could I ask one more thing of you... that you keep the tests to JUnit for now? I'm not against switching over to Spock, but I'd prefer not to have some tests in one testing framework and the rest in another. 

If Spock is the better choice then lets create a separate issue for the conversion of the whole test suite over to Spock.

ancho:
No problem @jonbullock 

Spock is such a clean, nice and concise testing framework, I don't even think about whether to use it or not, I just do if I can. 
It enables you to write really readable and understandable tests. The build-in mocking and stubbing support is really fun to use and the data centric test support is a dream.

I can backport it to junit and use mockito. Shouldn't be that hard. And maybe it is a good demonstration to see the two different approaches.

jonbullock:
Amazing work thanks @ancho

PR #293, opened 2016-06-25
Implemented Categories Support

manikmagar:
Implemented Categories Support for Posts.

### Categories configuration controls - 

- `template.category.file=category.ftl` _(This is for rendering individual category page)_
- `template.categories.file=categories.ftl` _(This is for rendering all categories list page to prevent directory listing on {site.host}/categories/ url._
- `categories.enable=true`
- `render.categories=true`
- `category.default=Uncategorized`
- `categories.path=categories`
- `categories.sanitize=false`

### Adding categories in content
Add an attribute `categories=Category1, Category2`
- If `categories.enable` is not false and no categories are provided for post then `category.default` is applied to post.
- First category is also available with `primary_category` key in model.
- If `categories.enable` is false and categories are still provided in content, then those are ignored with a warning in log.

### Accessing categories in templates
- `all_categories` - All categories from all published content. This variable is available on templates.
- `category_posts` - On a single category page `template.category.file`, this variable is available and provides access to posts marked with category being displayed on page.

### Extensionless Support
If `uri.noExtension` is set to `true` then all category pages are rendered without extension.

### Test cases
Appropriate test cases have been added under `src/test/java` with its resources under `src/test/resources`.

### DIFF
--- a/src/main/java/org/jbake/app/ContentStore.java
+++ b/src/main/java/org/jbake/app/ContentStore.java
@@ -133,6 +133,10 @@ public class ContentStore {
         return query("select * from post where status='published' and ? in tags order by date desc", tag);
     }
 
+    public List<ODocument> getPublishedPostsByCategories(String category) {
+        return query("select * from post where status='published' and ? in categories order by date desc", category);
+    }
+    
     public List<ODocument> getPublishedPages() {
         return getPublishedContent("page");
     }
@@ -153,6 +157,10 @@ public class ContentStore {
         return query(query + " order by date desc");
     }
 
+    public List<ODocument> getAllCategoriesFromPublishedPosts() {
+        return query("select categories from post where status='published'");
+    }
+    
     public List<ODocument> getAllTagsFromPublishedPosts() {
         return query("select tags from post where status='published'");
     }
@@ -203,6 +211,16 @@ public class ContentStore {
 	    for (ODocument document : docs) {
 	        String[] tags = DBUtil.toStringArray(document.field("tags"));
 	        Collections.addAll(result, tags);
+	    }
+	    return result;
+    }
+    
+    public Set<String> getCategories() {
+    	List<ODocument> docs = this.getAllCategoriesFromPublishedPosts();
+	    Set<String> result = new HashSet<String>();
+	    for (ODocument document : docs) {
+	        String[] categories = DBUtil.toStringArray(document.field("categories"));
+	        Collections.addAll(result, categories);
 	    }
 	    return result;
     }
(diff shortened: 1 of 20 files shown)


jonbullock:
Thanks for taking the time to submit this @manikmagar I'll review this properly once v2.5.0 is out and planning starts for the next release.

manikmagar:
@jonbullock Any date for 3.0 release? This is one of the few things that's either requiring me to use my own version of jbake or just skip categories and use tags (not so much good option). I would prefer to use standard jbake version :) Thanks.

jonbullock:
No date planned at present. However I'm learning towards having a 2.7.0 release, especially as some big changes are lined up for 3.0.0 and this would be an ideal candidate for another 2.x release.

manikmagar:
Thank you @jonbullock , I will eagerly be waiting for 2.7.0 :+1:

manikmagar:
This was implemented a while back (on maven), I think I will rebase to current master and update it.

jonbullock:
It would be great if you could thanks @manikmagar

src/main/java/org/jbake/parser/MarkupEngine.java
```diff
@@ -140,6 +141,31 @@ public void processBody(final ParserContext context) {}
...
+	                	categories[i]=categories[i].replace(" ", "-");
+	                }
+	            }
+	            content.put("categories", categories);
+	        } else {
+        		String defaultCategory = config.getString(Keys.CATEGORY_DEFAULT).trim();
+        		if (config.getBoolean(Keys.CATEGORY_SANITIZE,false)) {
+        			defaultCategory = defaultCategory.replace(" ", "-");
+                }
+        		content.put("categories", ArrayUtils.toArray(defaultCategory));
```
GeorgHenkel: Also here it would be better to use Crawler.Attributes.CATEGORIES instead of the string value "categories"

src/main/java/org/jbake/parser/MarkupEngine.java
```diff
@@ -140,6 +141,31 @@ public void processBody(final ParserContext context) {}
...
+	        }
+			content.put("primary_category", ((String[])content.get("categories"))[0]);
+        } else {
+        	if (content.get("categories") != null) {
```
GeorgHenkel: Also here it would be better to use Crawler.Attributes.CATEGORIES instead of the string value "categories"

src/main/java/org/jbake/app/ContentStore.java
```diff
@@ -242,6 +258,16 @@ private void executeCommand(String query, Object... args) {
         return result;
     }
 
+    public Set<String> getCategories() {
+    	DocumentList docs = this.getAllCategoriesFromPublishedPosts();
+	    Set<String> result = new HashSet<String>();
+	    for (Map<String, Object> document : docs) {
+	        String[] categories = DBUtil.toStringArray(document.get("categories"));
```
manikmagar: Good catch @GeorgHenkel, Thanks. I will make the changes.

PR #294, opened 2016-06-26
Permalink Generation and customization for content

manikmagar:
This pull request implements Permalink/URI Generation and customization in JBake.

It is driven by a property `permalink` in jbake.properties file. Sample permalink pattern can be `permalink=/:blog:/:title` which will generate urls as `blog\first-post.html`.

If `permalink` is not specified then default pattern is `/:filepath`.

Every keyword should be separated with '/:'. It allows you to use **any keywords** from content header. eg. `/:slug/:tags/:title`.

### Few Rules:
Some keywords are reserved for specific format. Below is the list of such keywords - 

1. **`/:filepath`**: This keyword will result it addition of relative source file path from `content` folder. If you want to maintain the content folder structure in output then, `permalink=/:filepath` will result in generating URI structure same as source files.
2. **`/:filename`**: This keyword will result in addition of source file name (without extension).
3. **`/:YEAR`, `/:MONTH`, `:DAY`**: These keywords will add a related date part from content published date to url.
4. **Adding Static words** - Static words can be added in url by using ':' as suffix in permalink. For example, `permalink=/:blog:/:filename` will result in generation of urls like /blog/First-post.html

### Using permalink in templates
Generated permalinks can be accessed in templates as `content.permalink`.

### Extension-less permalinks
If `ur.noExtension` is set to `true` then all permalinks will be extension-less. For example, `permalink=/:blog:/:filename` will generate `/blog/First-post/`.

### How to use different permalink patterns for each content type.
It is possible to use different permalink patterns based on content type. For example, post and page content type can use their own permalink patterns. `permalink` property will always be used as a default pattern for every content. To customize it based on content type, you can add another property `permalink.{type}` and specify pattern for it.

Sample configuration -
`permalink=/:filepath`
`permalink.post=/:blog:/:filename`
`permalink.page=/:filename`

Above configuration will result in pages with uri as filenames only, while posts will have suffix /blog/ for url. Any other content type will keep using `/:filepath` pattern.


jonbullock:
Interesting idea... I'm interested to understand what drove you to make this change? What is the use case you wanted to achieve that you couldn't without the change? As the output is driven by the source structure.

manikmagar:
@jonbullock , Sorry for the late reply, but I remember I was migrating my wordpress site where all posts got exported to root of the folder but in wordpress they had different url pattern. So instead of putting them into folder structure, I thought there should be a way to define permalink pattern. Hence I added this one :).

jonbullock:
Thanks for providing the context behind this @manikmagar it has certainly helped 👍

jonbullock:
👍

PR #296, opened 2016-07-06
shutdown orient database after bake has finished.

ancho:
Using a local storage cache with `db.store=local` throws an exception
rerunning the bake task.

```
:bake FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':bake'.
> Cannot open local storage 'cache' with mode=rw
```

See [stacktrace](https://gist.github.com/ancho/a031859dd4a04f497c91a12a51c09c7f).

### DIFF
--- a/src/main/java/org/jbake/app/Oven.java
+++ b/src/main/java/org/jbake/app/Oven.java
@@ -1,25 +1,18 @@
 package org.jbake.app;
 
-import java.io.File;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.ServiceLoader;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import com.orientechnologies.orient.core.record.impl.ODocument;
-
 import org.apache.commons.configuration.CompositeConfiguration;
-import org.apache.commons.configuration.ConfigurationException;
 import org.jbake.app.ConfigUtil.Keys;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jbake/app/Oven.java
```diff
@@ -174,7 +167,7 @@ public void bake() {
                 }
         } finally {
                 db.close();
-//                Orient.instance().shutdown();
```
jonbullock: This was removed because it caused an issue while running in JBake's watch mode... but the cause of that may have been fixed since.... could you check if this still breaks watch mode?

ancho: Ah ok. Yes I can do that. I'm using jbake in combination with gradle and the continous task runner. Therefor I'm not aware of this breaking the watch mode for the cli.

ancho:
Hey Jon, yes it did break. When an instance of the Orient class gets created it calls ``startup()`` on the first instantiation. Calling ``shutdown()`` does a lot of stuff, like unregister engines.
The easiest solution is to check if count of engines is zero and call ``Orient.getInstance().startup()`` to reinitialize the orient database. 
It works with watch mode for the cli and with a continuous gradle task.

On the long term it is better to decouple the startup and shutdown of the database from the ``bake`` method by passing the ``ContentStore`` to the ``Oven`` and handle it elsewhere I guess.

jonbullock:
Thanks @ancho and I agree this needs to be looked at long term.

PR #298, opened 2016-07-24
Add tests for DocumentsRenderer

ancho:
Added some Tests for `DocumentsRenderer`.
Refactored `ContentStore` to use `DocumentList` as query results.
Removed and replaced `DocumentIterator` in favour of the `DocumentList` abstraction.


jonbullock:
Always like to see coverage going up 👍 

Thanks @ancho

PR #304, opened 2016-09-24
Update Apache Commons IO from 2.4 to 2.5

obfischer:


jonbullock:
I usually perform a review of dependencies prior to each release in terms of upgrades but I'll pull this one in now anyway.

Thanks again.

PR #305, opened 2016-09-24
Update of AssertJ from 1.7.1 to 3.5.2

obfischer:


obfischer:
Very version of AssertJ higher then 1.7.1 requires at least Java 7.

PR #309, opened 2016-10-18
Ignore javadoc error on package

lamvak:
Running
mvn clean package
[Log from master build](https://gist.github.com/lamvak/6726d6e07786755ec7c09b02bac259e8)
[Log from ignore_javadoc branch build](https://gist.github.com/lamvak/9f1b828c40f82c4cf95b2d56608e267a)

Of course, since javadoc is used for documentation, it would be better to just fix the issues. However this allows to build the package with current shape.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,6 +121,9 @@
 	        </plugin>
 	        <plugin>
 	          <artifactId>maven-javadoc-plugin</artifactId>
+                <configuration>
+                    <additionalparam>-Xdoclint:none</additionalparam>
+                </configuration>
 	          <executions>
 	            <execution>
 	              <id>attach-javadocs</id>


jonbullock:
I suspect this is down to JDK8 and it's doclint, could you try the fix suggested in this blog - disabling it on the plugin section of the POM: http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html

If it works for you could you update your PR with this change to replace the "failOnError" - thanks.

lamvak:
All seem good now: [see new log file](https://gist.github.com/lamvak/6726d6e07786755ec7c09b02bac259e8#file-disable_doclint_build-log)
I propose to merge this as tactical fix for Java 8 compatibility and when I have a moment I can try to make the javadoc compatible with both Java 8 and 7

jonbullock:
Excellent. I'll merge this in now and be happy to accept the updated javadoc. 👍

PR #312, opened 2016-10-28
fixes #310 - add nextFileName and previousFileName to the model

### LINKED ISSUE #310: Index variables don‘t exist despite enabled pagination
### Problem

I used the latest JBake 2.5.0 release, enabled paging and set the post per page parameter:

```
index.paginate=true
index.posts_per_page=10
```

However in my freemarker template the variables `nextFileName` and `previousFileName` does not exist:

```
<#if (nextFileName)??> <#-- is always false -->
   [...]
</#if>
<#if (previousFileName)??> <#-- is always false -->
   [...]
</#if>
```
### Possible solution

In `org.jbake.app.Renderer#renderIndex()` the variables are put to the model map:

```
[…]
model.put("previousFileName" …);
[…]
model.put("nextFileName",  …);
[…]
```

 But the model is never given to freemarkers template engine. Maybe here the `ModelRenderingConfig` is needed instead of the used `DefaultRenderingConfig`.

ancho:
- adds `currentPageNumber` to the model to get the current page number
- adds `numberOfPages` to the model to get the total number of pages

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
 	<groupId>org.jbake</groupId>
 	<artifactId>jbake-core</artifactId>
-	<version>2.5.0</version>
+	<version>2.6.0-SNAPSHOT</version>
 	<packaging>jar</packaging>
 
     <name>jbake</name>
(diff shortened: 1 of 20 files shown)


jonbullock:
I just left a comment on #313 asking if you could separate out the fix for #310 but then noticed this PR. Is this the fix for #310 ?

ancho:
right. I'm confused, too. Sorry :)
#312 is the fix for #310 and #313 is on top of it. Phew...

jonbullock:
OK gotcha. We've got there in the end :)

Issue #314, opened 2016-11-04
JBake JettyServer doesn't work with custom destination


jmcgarr:
I am running a JBake site using the Gradle plugin. I am partial (and so it the plugin) to using the `src/jbake` directory for source and the `build/jbake` directory for output. I cannot get jbake to serve content using these two directories. Here's the command I used:

`jbake src/jbake build/jbake -b -s`

This results in a successful bake to the `build/jbake` directory, but the Jetty server presents content from the `output` directory. 

Planning on trying to create a unit test to replicate this.

jonbullock:
At first glance this is probably down to the fact that the -s option is configured to use the source parameter.

ancho:
It is also possible to apply the gradle gretty plugin. For example:

```
plugins {
     id 'org.akhikhl.gretty' version '1.4.0'
}

 gretty {
     httpPort = 8881
     contextPath = "/blog/"
     extraResourceBases = ["build/jbake"]
 }
```

Run `gradle jettyRun` to fire up jetty with your fresh baked website.

See https://github.com/ancho/calmdev-website for a configuration with different target environments (local, test and prod).

jmcgarr:
Thanks for that @ancho. 

I really wanted the `jbake-gradle-plugin` to take advantage of the new file watching features built into the `Main.java`. My hope is to land this fix so I can push that PR as well.

ancho:
gradle has a continuous build feature `gradle -t`. It watches for changes of the task inputs an runs the task again if something has changed.

See https://www.calmdevelopment.de/blog/2016/08/jbake-continuous-with-gradle.html

ancho:
But don't get me wrong. It's still a bug and need to be fixed. Thank you for cleaning up and improving the application.

jonbullock:
See #316

PR #316, opened 2016-11-06
Improve testability of Main.java while fixing issue #314

jmcgarr:
My goal for this pull request is to allow the `jbake/jbake-gradle-plugin` an API to start the `JettyServer`. While adding this functionality to the jbake-gradle-plugin, I observed a bug (#314) that prevents the `JettyServer` to start up with a custom output directory as well as providing the watcher with a custom source directory. The fix for this issue was fairly straightforward.

While adding this change to the `Main.run()` method, I recognized an opportunity to refactor the code to make it more testable. This allowed me to write tests for the failure cases I detected and provide fixes for them.

### DIFF
--- a/src/test/java/org/jbake/launcher/MainTest.java
+++ b/src/test/java/org/jbake/launcher/MainTest.java
@@ -1,0 +1,105 @@
+package org.jbake.launcher;
+
+import org.apache.commons.configuration.CompositeConfiguration;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.kohsuke.args4j.CmdLineException;
+import org.kohsuke.args4j.CmdLineParser;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+import static org.mockito.Mockito.verify;
+
+@RunWith( MockitoJUnitRunner.class )
+public class MainTest {
+
+    private Main main;
+
+    @Mock private Baker mockBaker;
+    @Mock private JettyServer mockJetty;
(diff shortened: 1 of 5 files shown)


jonbullock:
Just been reviewing this change and I'm assuming this breaks the execution of the `-s` option when it's used on it's own with just a single path parameter?

jmcgarr:
It shouldn't break this and there is a test case to make sure that option still works.

src/test/java/org/jbake/launcher/MainTest.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    // Documentation states these two commands will define the custom output, but the LaunchOptions file isn't setup for that.
+    // I have written this test to define the existing functionality of the code and not that defined in docs.
+    @Test
+    public void launchJettyWithCustomDestinationDir() {
+        String[] args = {"-s", "build/jbake"};
+        main.run(args);
+
+        verify(mockJetty).run("output","8820");
```
jonbullock: This is the scenario I was referring to in my previous comment, in v2.5.0 the result of this execution `jbake -s build/jbake` would be that the contents of `build/jbake` would be served out not `output`. The underlying issue is the sharing of the source & destination values for different cmd line options.

jonbullock:
I've added a line comment for the scenario I was referring to and how 2.5.0 currently behaves. While I don't want to break existing functionality or at least the way JBake behaved, I do want to merge this in as it improves the testability of the launch commands. I'll do a bit more research and see if I can come up with a quick fix until a better overhaul of the command line interface can be implemented.

jonbullock:
I've added a workaround to maintain the existing behaviour while still allowing the Jetty server to serve out the destination path when combined with the bake option and explicit source & destination locations.

Thanks @jmcgarr for submitting this change 👍

Issue #266, opened 2016-02-10
"alltags" does not contain tags from custom template types


mkobit:
I was looking to use JBake and Freemarker to generate a static site and use my own custom templates for rendering certain pages, rather than using `post` and `page` as those do not have semantic maening in the site I am generating. 

The query `query("select tags from post where status='published'")` is run when trying to retrieve `"alltags"`, and from what I have discovered a table is created for each `"docType"`. It was a little strange to me that the "global" values are not necessarily collected from all types.

oliverlietz:
@mkobit I'm interested in this _feature_ also. Do you have a patch to take tags from all types into account?

jonbullock:
Thanks for raising this issue.

PR #323, opened 2016-12-12
Tag pages are created for documents of type post only

oliverlietz:
`org.jbake.app.Renderer#renderTags(String)` calls `ContentStore#getTags()` where it should call `ContentStore#getAllTags()` (IMHO), see #266 @jonbullock, WDYT?


jonbullock:
Yeah @oliverlietz you're spot on, thanks for catching this one.

PR #328, opened 2017-01-11
Use StringBuilder more consistently

ottlinger:
* Replace StringBuffer with StringBuilder
* use append() instead of concatentaion


jonbullock:
Thanks for the contribution.

Issue #333, opened 2017-01-14
jbake -s should automatically override site.host


johnament:
The property `site.host` is defined in `jbake.properties` to rewrite URLs.  This works, except when you're using `jbake -s` to run your site.  This is running on `localhost:8820` however generated links that reference `config.site_host` point to whatever is in `site.host`

with this enhancement, jbake should know that its running with -s and use the local service URL instead of what is in `site.host`

PR #334, opened 2017-01-14
Enhancement for #333, override site.host variable when running the server.

johnament:
This is an enhancement for #333 so that when you do a -s (or a jbake -b -s) the value of site.host is automatically populated for the running server, to avoid overloading it.

Another thing added in here is that system properties will be brought in by default now.  This would allow a user to override any settings they may have from the CLI.

I also did a little bit of testing clean up in this.


jonbullock:
Thanks for the enhancement.

PR #337, opened 2017-01-24
update orientdb to 2.2.15

### LINKED ISSUE #318: update orientdb
A user at the mailing list was confused about the warning "Current implementation of storage does not support sbtree collections" . See https://groups.google.com/forum/#!topic/jbake-user/8hdkh-zl_Sk

It has maybe something to do with indexes and the currently used version of orientdb (1.7.10).

The current version of orientdb is 2.2.12

ancho:
* added -XX:MaxDirectMemorySize=1024m to jbake and jbake.bat script, to prevent warning from orientdb

closes #318

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,7 @@
         <junit.version>4.12</junit.version>
         <pegdown.version>1.6.0</pegdown.version>
         <jetty.version>8.1.19.v20160209</jetty.version>
-        <orientdb.version>1.7.10</orientdb.version>
+        <orientdb.version>2.2.15</orientdb.version>
         <groovy.version>2.4.7</groovy.version>
         <slf4j.version>1.7.21</slf4j.version>
         <logback.version>1.1.7</logback.version>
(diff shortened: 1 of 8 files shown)


ancho:
The openjdk6 build fails as orientdb [requires java 7 or higher](http://orientdb.com/docs/last/Tutorial-Installation.html#prerequisites). It's time to drop java 1.6 support I guess.

jonbullock:
I'll mark this for 2.6.0 but I'm torn whether a bumping the JRE required level is considered a major or minor change....

ancho:
Hmm. That's a very good question. Semantic versioning for me is all about API changes.
Is a runtime requirement a API change or not? It's kind of a grey area.
But thinking in terms of the definition of major and minor. It is more a major change than a minor change.
On the other hand, neither a developer nor a user should use java 6 in 2017.
So a runtime requirement could also be possible to change in a minor version.
Needs a bit research on how other projects decided on that question.

ottlinger:
+1
I'd prefer 2.6.0 and an explicit hint in the developer documentation and release notes

jonbullock:
I'll leave it for 2.6.0 for the time being and review closer to the time.

jonbullock:
Plan is to switch to Java 7 for 2.6.0+ (and Java 8 for 3.0.0+) - https://groups.google.com/d/msg/jbake-dev/01SO2OBJ8VA/2IuIiEf9CAAJ

Issue #340, opened 2017-02-04
Introduce a configuration abstraction to access configuration properties in a clean and concise way


ancho:
This is more a cleanup and refactoring Task. See comments from #300.

JBake uses [apache commons CompositeConfiguration](https://commons.apache.org/proper/commons-configuration/userguide/howto_compositeconfiguration.html) or it's interface `Configuration` to access configuration properties.

It's all over the place and accessed in the same way, with some constant keys defined in `ConfigUtil.Keys` or variable ones that may be added to the jbake.properties file like custom types.

It would be nice to have a configuration abstraction that is intialized during a configuration phase and shared between classes, validating requried properties like a source or destination folder and things like that. See a [pseudo implementation](https://gist.github.com/ancho/f61643f93d6bdb9a73c90dca32202d3e) of a JbakeConfiguration as an example of how it could look like.

Here is a list of Classes that make use of a CompositeConfiguration. Some of them take parameters that should be already present in the configuration.

* Asset
  * `public Asset(File source, File destination, CompositeConfiguration config)`
* Crawler
  * `public Crawler(ContentStore db, File source, CompositeConfiguration config)`
* Oven
  * `public Oven(final File source, final File destination, final CompositeConfiguration config, final boolean isClearCache)`
* Renderer
  * `public Renderer(ContentStore db, File destination, File templatesPath, CompositeConfiguration config)`
* Init
  * `public Init(CompositeConfiguration config)`
* CustomFSChangeListener
  * `public CustomFSChangeListener(LaunchOptions res, CompositeConfiguration config)`
  Uses the configuration to instantiate an Oven and pass the configuratin to the constructor

* AbstractTemplateEngine 
  * `protected AbstractTemplateEngine(final Configuration config, final ContentStore db, final File destination, final File templatesPath)`
  * extended by
    * `DelegatingTemplateEngine`
    * `FreemarkerTemplateEngine`
    * `GroovyMarkupTemplateEngine`
    * `GroovyTemplateEngine`
    * `JadeTemplateEngine`
    * `ThymeleafTemplateEngine`

And used in different utility classes.

* Baker
  * `public void bake(final LaunchOptions options, final CompositeConfiguration config)`
  Instanciates an Oven and passes the config to the constructor
* FileUtil
  * `public static String findExtension(CompositeConfiguration config, String docType)`
  Extract a fileextension from configuration by document type
* BakeWatcher
  * `public void start(final LaunchOptions res, CompositeConfiguration config)`
[...]

PR #341, opened 2017-02-09
refactoring of jbake configuration

ancho:
I started to refactor the configuration of jbake and it's components. See #340.

In general it's finished. But I want to write some more tests and add some logging in case options could not be found and things like that.

There are only two classes left with a direct dependency to a CompositeConfiguration, which are configuration specific classes.

Please take your time to review the changes. I hope you like it.


ancho:
Ok. I think I'm finished.

ancho:
Hello? Anybody out there? :)

jonbullock:
Still here.... just busy preparing for a talk at the moment :)

ancho:
All right. Just a ping. Worked obviously :)
Good luck with the talk then.

jonbullock:
I haven't forgotten about this, I'll try and catch you on IRC to discuss what order is best to merge this and the other major PR's (such as the switch to Gradle).

ancho:
Hmm...I'm rebasing from time to time and I ask my self if it would find a way earlier into master if I add deprecated versions of the old Method/Constructor signatures that changed during the refactoring?

jonbullock:
Thanks for rebasing this and keeping it up to date with master 👍 

I've scheduled in a 2.7.0 release now so if you can add those deprecated methods/constructors for the build plugins (anything else as well?) I don't see why we can't include this in that release. What do you think?

ancho:
I'll try to keep it up to date. It would be nice to see this as soon as possible. But I'm not sure if it's possible to get a 100% compatible version. The [FileUtil](https://github.com/jbake-org/jbake/pull/341/files#diff-80cc01ddb39611b1590bf970b26c5768) lost the public static `findExtension` for example. But it doesn't make sense to add this method there. It's the wrong place and is part of the [DefaultJBakeConfiguration](https://github.com/ancho/jbake/blob/3d55f33e2c598118636f0fdad09a9eee56d06b33/src/main/java/org/jbake/app/configuration/DefaultJBakeConfiguration.java#L83) now.

I'll open a new branch for this experiment and concentrate on the main classes like

* Crawler
* Parser
* Asset
* Oven
* ContentStore
* Renderer
* implementations of MarkupEngine 
* implementations of RenderingTool
* etc.

I touched 69 files. Ok...I can substract the tests but that will take some time.
The question is what classes are API classes and which modifications count as a normal refactoring changes that don't violate semantic versioning.

ancho:
Ok. I created a new [branch](https://github.com/ancho/jbake/tree/feature/configuration-refactoring-compatibility) with restored methods and constructors.

I tested the snapshot version against the gradle plugin and my website project and it still works.

The CI failed a few times before, because the singleton ModelExtractors had registered engines from previous tests. So I had to implement a reset method to keep the tests clean and seperated from each other.
The same applies for DocumentTypes which already had a rest method.

I'm not quite sure if this is a problem for production too...

jonbullock:
I can test the branch against the Maven plugin as I've just published v0.2.0 if you want to see if there is any major problems? If not we can schedule it in for 2.7.0.

ancho:
That would be great. I will try to implement a test to verify we do not get problems with the singletons running bake in a loop removing document types. The only scenario I can think of at the moment.

ancho:
Rebased to master and squashed to make a review easier.

jbake-core/src/main/java/org/jbake/app/configuration/ConfigUtil.java
```diff
@@ -0,0 +1,59 @@
...
+            config.addConfiguration(new PropertiesConfiguration(customConfigFile));
+        }
+        config.addConfiguration(new PropertiesConfiguration(DEFAULT_CONFIG_FILE));
+        config.addConfiguration(new SystemConfiguration());
+        return config;
+    }
+
+    private void displayLegacyConfigFileWarningIfRequired() {
+        LOGGER.warn("You have defined a part of your JBake configuration in {}", LEGACY_CONFIG_FILE);
+        LOGGER.warn("Usage of this file is being deprecated, please rename this file to: {} to remove this warning", CONFIG_FILE);
```
manikmagar: Should we say that `Usage of this file is deprecated and will be removed in JBake 3.0. Please rename this file to {}`?

jbake-core/src/main/java/org/jbake/parser/AsciidoctorEngine.java
```diff
@@ -89,40 +89,57 @@ public void processHeader(final ParserContext context) {
...
+                String dateFormat = context.getConfig().getDateFormat();
+                DateFormat df = new SimpleDateFormat(dateFormat);
+                try {
+                    Date date = df.parse((String) value);
+                    contents.put("date", date);
+                } catch (ParseException e) {
+                    LOGGER.error("Unable to parse revdate. Expected {}", dateFormat, e);
                 }
             }
             if (key.equals("jbake-tags")) {
```
manikmagar: Should we extract "jbake-tags" to a constant?

ancho: Yes. There is a lot of stuff we should put into constants or find better substitutes.
I just wanted to focus on the configuration part with this PR.

ancho:
rebased to master

ancho:
rebaseCount++

jonbullock:
Shall we merge this before we merge smaller PR's like #471 ?

ancho:
Makes sense. That way we "just" need to adjust the outstanding PR's.

jonbullock:
Do you want to do the honours for this one?

ancho:
My pleasure! Do you need to prepare something before? Should I wait or just merge it right now?

jonbullock:
No I'm OK, go for it.

PR #358, opened 2017-05-10
Fixed Issue#357: Accessing db in freemarker template throws exception

manikmagar:
Fix for #357 and unit test case for it.


jonbullock:
Thanks for taking the time to submit a fix for this bug 👍

Issue #360, opened 2017-05-13
Relative path's are broken due to wrong rootpath when extensionless uri is enabled.


manikmagar:
When extensionless uri's are enabled. [Crawler](https://github.com/jbake-org/jbake/blob/master/src/main/java/org/jbake/app/Crawler.java) is making rootpath first but not updating it with one more '../' when extension less uri is on.

jonbullock:
Thanks for catching this one.

manikmagar:
Working on a fix, will send PR soon.

manikmagar:
I found same issue for paginated index pages also. eg. blog/2/index.html was missing one level "../" in rootpath. PR fixes both issues.

jonbullock:
👍

manikmagar:
Do you have a date in mind for 2.6? I feel the bug fixes should be released sooner than enhancements. May be 2.5.2!

jonbullock:
The date I originally had in mind has come and gone, so it's ASAP at the moment. I'll review what's left to do for 2.6.0 and if it's a while off I'll look at pushing out a bug fix release.

manikmagar:
This bug has left the extensionless uri and pagination not usable as all static asset references using relative paths will break causing sites not to render properly. 

**Workaround in meantime:**
For sites rendered at root, absolute paths can be use eg. /css/main.css instead of relative like ../css/main.css.
For subfolder sites, add folder name /blog/css/main.css

benelog:
I am waiting release of 2.6.0 for a month.
If the 2.6.0 release takes time, it may be better to release 2.5.2 to fix this issue.

PR #361, opened 2017-05-14
Corrected Rootpath for extension less uri and paginated indices.

manikmagar:
Fixes Issue #360 for Extensionless URI and paginated index pages.


jonbullock:
Thanks for taking the time to fix this.

PR #362, opened 2017-05-14
Exclude index.html from previousFileName and nextFileName values.

manikmagar:
When index is paginated, previousFileName and nextFileName values are
added to content model. These values were including index.html in it.
For example, nextFileName value on first page was "2/index.html". It is
better to have values just as "2/" or "3/" etc. Corresponding test cases
have also been updated.

@ancho @jonbullock, Please take a look at this.

### DIFF
--- a/src/main/java/org/jbake/util/PagingHelper.java
+++ b/src/main/java/org/jbake/util/PagingHelper.java
@@ -17,7 +17,7 @@ public class PagingHelper {
 
     public String getNextFileName(int currentPageNumber, String fileName) {
         if (currentPageNumber < getNumberOfPages()) {
-            return (currentPageNumber + 1) + File.separator + fileName;
+            return (currentPageNumber + 1) + File.separator;
         } else {
             return null;
         }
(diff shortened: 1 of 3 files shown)


ancho:
Thank you. We maybe should change the [index.ftl](https://github.com/jbake-org/jbake/blob/master/src/test/resources/fixture/freemarkerTemplates/index.ftl#L19) template in the fixture to reflect those changes as well. And change the [test](https://github.com/jbake-org/jbake/pull/362/files#diff-6d1a820f8dbc696c796be269c47fa146R53) to expect the root path in the previous link.
What do you think?

manikmagar:
I did update the test cases to expect the root,changes are in this PR. Index.ftl should be fine but in implementation, it needs to prepended with content.rootpath. See in my this [template code](https://github.com/manikmagar/jbake-future-imperfect-template/blob/master/templates/post/prev-next.ftl). If you bake that template with master copy of source, you can see it. Also I made another PR which corrects the rootpath on pagination indices.

jonbullock:
Thanks for this @manikmagar

PR #363, opened 2017-05-14
add tests for built in example projects

ancho:


jonbullock:
Excellent improvement... saves me a job when doing creating a release 👍

ancho:
That's nice. You maybe better take a look at #366. It includes this one with fixes for windows.
I had a bug in the test which caused the processes to hang on windows.

PR #367, opened 2017-05-17
Prevent image source path from breaking

manikmagar:
This change prevents image source path from breaking. It prefixes
content rootpath to the image sources. 

It excludes the sources that starts with http:// or https:// or /.

If image source starts with './' i.e. image is in the same folder as
content, then it will replace ./ with content directory path and then
adds the rootpath.

Once image source is fixed, it should not break, no matter from where
content is accessed i.e. index or individual page.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -46,6 +46,7 @@ ext {
     jsonSimpleVersion = '1.1.1'
     jade4jVersion = '1.2.5'
     mockitoVersion = '1.10.19'
+    jsoupVersion = '1.10.2'
 }
 
 dependencies {
@@ -68,6 +69,7 @@ dependencies {
     compile group: 'ch.qos.logback', name: 'logback-classic', version: logbackVersion
     compile group: 'ch.qos.logback', name: 'logback-core', version: logbackVersion
     compile group: 'de.neuland-bfi', name: 'jade4j', version: jade4jVersion
+	compile group: 'org.jsoup', name:'jsoup', version: jsoupVersion
     testCompile group: 'junit', name: 'junit', version: junitVersion
     testCompile group: 'org.assertj', name: 'assertj-core', version: assertjCoreVersion
     testCompile group: 'org.mockito', name: 'mockito-core', version: mockitoVersion
--- a/src/main/java/org/jbake/app/Crawler.java
+++ b/src/main/java/org/jbake/app/Crawler.java
@@ -13,6 +13,7 @@ import org.jbake.app.Crawler.Attributes.Status;
 import org.jbake.model.DocumentAttributes;
 import org.jbake.model.DocumentStatus;
 import org.jbake.model.DocumentTypes;
+import org.jbake.util.HtmlUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -180,7 +181,7 @@ public class Crawler {
             
             
             // Prevent image source url's from breaking
-            HtmlUtil.fixImageSourceUrls(fileContents);
+            HtmlUtil.fixImageSourceUrls(fileContents,config);
 
             ODocument doc = new ODocument(documentType);
             doc.fields(fileContents);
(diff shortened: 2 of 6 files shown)


jonbullock:
Ah excellent, this addresses a common issue right now.

manikmagar:
Although this address image source issue while rendering posts/pages on the site, 
I think still the image path in rss feed will be broken when viewed on external feed reader. Image sources will be relative. 

@jonbullock do you think, we should make them fully qualified path instead of relative? eg. http://example.com/content/a.jpg instead or content/a.jpg?

jonbullock:
@manikmagar Yeah I agree we should make them fully qualified for the feed for this release. We can review this in subsequent releases.

manikmagar:
@jonbullock Feed just loads the content already parsed into HTML source code. To make feed paths working, we have two options - 

1. Use FQDN for everything, this will work when site is deployed to server but for local development, everytime jbake.properties will need to be updated to have localhost as site.host. Else, images will not load in local. I think, this might be inconvenient for development.
2. After Feed is rendered, read the generated XML file to correct image sources in description content. Not sure, if this option is good.

Do you have any thoughts or suggestion?

ancho:
👍 for full qualified domains. The [ConfigUtil](https://github.com/jbake-org/jbake/blob/master/src/main/java/org/jbake/app/ConfigUtil.java#L246) rewrites the site.host property when running with jetty. You need to run with -b option too. For a clean workflow, say running a server in background with -s while editing content, we would need a mechanism to detect a still running jetty instance launched from jbake to do the same thing.

manikmagar:
When file is parsed, the rendered html content is stored in `fileContents` with key `body`. How about after fixing the `body`, we add 'feedBody' with FQDN's in it. Feed templates which now use `body` can use `post.feedBody` -

``` <description>
	<#escape x as x?xml>	
	${post.feedBody}
	</#escape>
	</description>```

May be a simple fix? What do you guys think about this?

jonbullock:
I didn't fully understand the complexity of this when I suggested FQDN for the feed.

I'm undecided whether adding an additional variable is better or worse than post-processing the XML output to make image references use FQDN instead of just relative. Adding a new variable just for use in the feed seems overkill but post-processing a rendered file doesn't feel right either.

manikmagar:
I was looking at img [srcset](https://www.w3schools.com/tags/att_source_srcset.asp) attribute that was added with html5. It is used for the responsive image set, browser first tries to load image from srcset and if that is not available then it falls back to src attribute. I am thinking to add relative path in srcset and FQDN in src. Except feed, browser should always then use relative path and when feed is rendered in reader, FQDN would be used.

manikmagar:
So, srcset did not work as expected. As @ancho said (btw thanks for that), ConfigUtil rewrites the site.host property to localhost when ran with `-b -s` together. So I changed the code to use FQDN always :). 

In case of local baking, if `-b -s` is used, all image url's will point to localhost.

For releasing site, when just `-b` is used, site.host as specified in the jbake.properties will be used.

This should solve the feed problem too.

Eg. 
Check the post at the bottom - On this [feed](https://json-feed-viewer.herokuapp.com/feed?url=https%3A%2F%2Fmanikmagar.github.io%2Fjbake-future-imperfect-template%2Ffeed.json)

For site rendering - check [this](https://manikmagar.github.io/jbake-future-imperfect-template/4/) post (it is the same as in feed above)

jonbullock:
Thanks @manikmagar any chance you can have a look at why the tests have failed after your latest commit?

manikmagar:
Ah! Sure. I will check and fix it.

manikmagar:
Corrected JSOUP API method call to prevent empty body during test
execution. This would also prevent extra body element in parsed
fragment.

jonbullock:
Excellent thanks @manikmagar

jonbullock:
Linking mailing list discussion on this: https://groups.google.com/d/topic/jbake-user/64MiZRTl_3I/discussion

PR #369, opened 2017-05-21
update dependencies

ancho:
Update of dependencies based on ./gradlew dependencyUpdates 

I couldn't resist to lift jetty from 8 to 9 :)


ancho:
Wow. OK. That was to fast. assertj dropped jdk 7 in 3.x

PR #371, opened 2017-05-24
Add rendering support to generate /tags homepage and tag model

### LINKED ISSUE #350: Make tagged documents data available as global data.
As of now, only `alltags` is available globally and it only contains String list of tag names. For one of the blogs, on home page, I want to show  number of documents in each tag, which doesn't seem possible or easily achievable.  May be alltags can return list of Tag model where Tag object could be a map of `{ "name" : "test", "noOfDocuments": 30, page-url: "{name/sanitizedname}"}` or more elements.

I can write a script in template to iterate over `published_posts` or `published_content` and then get the stats but that will slow down rendering along with adding other scripting complexities to templates.

P.S. I actually wanted to show categories and their post count but until #293 is available in release, I am choosing tags. Although, even categories then should be made available globally with more data elements.

### LINKED ISSUE #351: Add rendering support to generate /tags homepage
Individual tags are rendered inside /tags or ${config.tags_path} folder but there is no way to define template page for /tags itself. As of now, if /tags is hit and if directory listing is enabled, then it just displays directory listing of tag html files. Instead of that, a new template should be added to allow rendering for /tags home page. This is helpful when you just want to add site menu 'Tags' which lands to all tags home page.  
eg.
`template.tag.home.file=tagshome.ftl`

For example, categories implemented in #293 .

Thanks.

manikmagar:
### Add rendering support to generate /tags homepage
A new property `template.tags.file=tags-index.ftl` is added. This
tags-index.ftl (or relavant remplate file) is rendered at the root of
tags path specified by `tags.path` property. Fixes #351

This also exposes a global `tags` data model which contains all
published tags along with tagged posts/documents. Fixes #350

### Corrected [model](https://github.com/jbake-org/jbake/commit/9a52ffae0a8144196a591f8cc2fac8c4e36e3abd#diff-7f997ee3aec4032e49234e831068c01fL77) set for jade rendering
JadeModel implements scopes for the model data. Earlier only the current
model was being passed which caused missing of config model. So changed
it to pass current instance of JadeModel which has get method
implemented to iterate through all scoped models. Earlier implementation
was causing model.get("config") to be null in TagExtractor for
JadeEngine but was working properly for other engines.

### [Allowed](https://github.com/jbake-org/jbake/commit/e782417ad2bd33f8d808a03efd8e41b38bca870f#diff-f3c2d91467207715566d7c7bc72fddddL30) to pull configuration values from parent, if set.


manikmagar:
FYI - This includes a fix for bug in Jade Template rendering, related to JadeModel.

Issue #274, opened 2016-03-27
jbakery - faster rebuilds and web console


ieugen:
Hello,

I've posted an email on the user mailing list about some improvements.
Right now thare are a lot initializations that can be skipped and are not. 
The code could make use of dependency injection to make the code easy to reuse. 

It is also my opinion that jbake currently does way more than it should in a single project"
- render pages
- start a webserver
- parse command line parameters

I think this should not be the case. I would split the code in multiple modules:
- jbake-core (or a similar name ) module that does rendering
- a jbake-cli (or jbakery) that uses the core and does command line parsing, web server and other stuff 

 Proof of concept project: https://github.com/netdava/jbakery

danielgrycman:
Hello @ieugen,

I share your opinion about multiple modules for jbake and I really like your poc :-).

ieugen:
Thanks @danielgrycman . 

I just added watch functionality to 0.0.4 ( you net to specify `--destination` outside  `--source`  or you get an infinie cycle atm) .
I'm working on adding ant style path matchers to ignore paths and files inside the work tree.

I have alot of ideas but a lot of them need jbake refactoring. Right now I'm still waiting for answers but I am considering forking the project. I would very much like to avoid that and that's why I'm coming with issues and questions.

jonbullock:
Just posted a reply to your message on the users mailing list. 

Watch mode has already been added to master: https://github.com/jbake-org/jbake/issues/253

The plan to split up the project into multiple modules is available here: https://github.com/jbake-org/jbake/issues/117 - are you interested in working on this?

If you want to start a discussion on the developers mailing list for your other ideas I can bring you up to speed with what is currently being worked on and by who?

ieugen:
Hi, 

Thanks for the reply. I agree that things should be discussed and this is how comunity will grow. 
Having that said, I did notice that those threads are 2 years old and jbake is still monolithic. 

I think its time for action. My take from the discussions is:
- a split is desired and needed. there are different ideas on how to do it but let's start and see how it goes.
- migrating to Gradle would help with that

I also think we should make the development/contribution process more open so the project will grow.  I say this because we are all too busy to contribute from time to time and it is desirable that the project moves forward regardless of our personal schedules.

I am a contributor to Apache Software Foundation, OPS4j, Debian Project and I know comunity is very important for an open source project. 

I would suggest jbake took some practices/policies/processes from those comunities.  

[...]

Issue #117, opened 2014-07-14
Move to sub-module structure


jonbullock:
As suggested here:

https://groups.google.com/d/msg/jbake-dev/e-hCPuZ4L6E/7UGoljG5ASMJ

jonbullock:
Reading material: http://www.petrikainulainen.net/programming/gradle/getting-started-with-gradle-creating-a-multi-project-build/

rajmahendra:
I can take this task if all are ok :) Maven or Gradle.

rajmahendra:
@jonbullock  We have around 15 pull request pending... If we have to proceed with this issue we have to first resolve all the pull requests ! https://github.com/jbake-org/jbake/pulls

lefou:
+1

jonbullock:
Seeing the AsciidoctorJ project move over to a sub-module Gradle build recently I thought we could do the same, so lets go with Gradle.

My plan was to start work on v3 in the v3.0.0 branch and merge in all the outstanding PR's into this branch while work carried on in master for v2.4.0. But having said that I don't know if this is possible since all the PR's are for the master branch?

lefou:
I would prefer to stay in master for v3 work and keep a 2.4 branch if required, but that's completely personal taste. In the end, it's only names.

Merging (at least some of) the pull request now before splitting the build would be easier of course. In best case no extra work needs to be done. If you modularize first, you need to apply all patches manually which is cumbersome and error-prone.

lefou:
Also, given that some PRs are older than a year, it's about time to merge or discuss and defer.

jonbullock:
Seems logical to me. I'll get v2.4.0 out the door as quick as possible then the PR's can be dealt with in preparation for the v3 release. If anything else needs doing on v2.4 a branch can be created for it.

Once the PR's are dealt with in whatever way required we can start to break the project down ready for v3.

rajmahendra:
I have created a demo subproject for jbake feel free to comment https://github.com/rajmahendra/jbakeproject

Did some changes in package name change. I yet to answer assemble.

jonbullock:
Thanks Raj, will definitely be useful when the break up into sub-modules begins.

PR #373, opened 2017-05-25
split into subprojects 

ancho:
This is an attempt to address #274.

There are 4 projects

* root aka. jbake-base
  * configures subprojects, jacoco execution aggregation and coveralls
* jbake-core
  * the core library. produces jbake-core-{version}.jar
  * publishes to bintray maven repository jbake-core
* jbake-cli
  * the cli library. produces jbake-cli-{version}.jar
  * dependency to jbake-core
  * publishes to bintray maven repository jbake-cli
* jbake-app
  * bundles the cli to an distribution
  * publishes to bintray binary repository jbake
  * publish to sdkman

While spillting the repository I noticed I forgot to publish the binary distribution to bintray.
It is way easier to do this with a separate application project. I had problems to configure the bintray plugin with a combined jar and distribution package to publish to different repositories.

You can run all tasks as you are used to.

If you want to run a task in a specific project from root run `./gradlew :jbake-cli:test` for example.

Now that I am writing this PR I think we should add an aggregated testreport to the root project to get a better overview.


ancho:
You need to remove the repo path in coveralls in order to see the source files.

jonbullock:
Linking to original idea for this #117 

The idea original was to separate out the code relating to each template engine supported and markup parsers. What's your thoughts on doing this?

manikmagar:
I think core should include the minimal set of template engines that will be supported out of the box. In addition to that, may be core allows to inject more template engines as plugin/extensions. So, if I want to use some custom template engine, I should not have to add it in core, I can just add an extension and use it. If there is a need to include that in core, due to more people using etc., we can then do that.

ancho:
I'm not sure about the structure yet. If the extension stuff is getting more real we should decide how to split into more modules that make sense. An api package with all the interfaces necessary to build extensions comes to mind. with core implementations, well...in the core package. And if we build the first extensions based on that api we could split them out as we need. So I would go the other path first. Adding new stuff to core and splitting it into modules only if necessary.
Or we split extensions out to a separate repository and make use of gradle [composite builds](https://docs.gradle.org/current/userguide/composite_builds.html) to keep the repository clean and maintainable.

jonbullock:
I believe the original idea for splitting out the template engines into separate modules was so you could selectively choose which to include as some come with some hefty dependencies - thinking of Groovy here. But it's not been an issue so far.

Probably a good idea to see how plugins/extensions will affect this.

ancho:
This was a proof of concept. The first step is done by splitting into core and distribution module in #422

Issue #327, opened 2017-01-11
Evaluate a replacement for pegdown markdown parser


ancho:
[pegdown](https://github.com/sirthias/pegdown) is marked deprecated. The author suggests [flexmark-java](https://github.com/vsch/flexmark-java).

jonbullock:
Thanks for raising this one.

PR #376, opened 2017-05-26
replace pegdown with felxmark-java

ancho:
See #327


jonbullock:
Thanks @ancho looks like this new lib delivers some serious performance gains.

bdelacretaz:
FWIW, for people experimenting with this on existing content, https://github.com/vsch/flexmark-java includes a "Pegdown Migration Helper" section.

kwin:
Would you reconsider integrating this really useful change into the next i.e. 2.6.0 version?

jonbullock:
Yeah we'll include this in the 2.6.0 release.

PR #384, opened 2017-06-16
Extended JBake watch mode to template and asset folder

### LINKED ISSUE #346: Extend watch mode to detect when changes are made to template folder
Linking related issue #343

manikmagar:
Closes #346


manikmagar:
@jonbullock check this please, if should be included in 2.6.

manikmagar:
@ancho do you mean using something like below instead of just printing stack trace? If yes, I don't see any issue with that, I can do that change.

`Logger.error('Failed to start file watcher. Changes to source, asset or template files will not be detected unless site is baked again.', e)`

jonbullock:
Looks good to me... I'll mark it for 2.6.0 now.

There are probably a few places I've just printed the stack trace out that need to be refactored to use the logging system.

Issue #381, opened 2017-06-09
Add support for plantuml content


giflw:
Would be nice be able to put plantuml files (.puml) on content folder and JBake invoke plantuml to generate the diagrams.

ancho:
If you are using gradle to bake your site you can include your .puml file in a plantuml block within an ascidoctor file to render them.
See https://www.calmdevelopment.de/blog/2016/09/jbake-asciidoctor-diagram.html

giflw:
I'm using just JBake. What's needed to make this? Is there some plugin support in JBake? I could try to implement something and make a PR.

ancho:
Not yet. There is some discussion going on and a prototype implementation of a future plugin system see https://groups.google.com/forum/#!topic/jbake-dev/jXu9oMQo5-4

If you are using asciidoctor you should be able to install the asciidoctor-diagram gem and use the same configuration I used in the gradle example withing your jbake.properties to do the same with the jbake cli. 
Use `gem install asciidoctor-diagram` to install the dependency.

But feel free to implement whatever comes into your mind. Maybe a good start to transform it to a plugin as soon as it gets real.

giflw:
I'll check those options.
Thanks
Best regards

ancho:
Ok. I was wrong with my assumption it should work with the gem. That makes no sense at all.
Sorry about that.

I want to add the asciidoctorj-diagram dependency to the distribution, so nobody has to add the dependency manually for cli and extend the CLASSPATH variable.

First step is #230. I'll open a PR soon...need to test a few things first.

But that does not solve the problem with images in content folders (#281). asciidoctor-diagram places the produced images in the same folder where the document resides in. Unless you configure the `imagesoutdir` attribute as a workaround which forces you to define the correct `imagesdir` attribute to get valid links.

Issue #230, opened 2015-09-29
Improve exec scripts to pick up all jar files in /lib


jonbullock:
Improve exec scripts to pick up all jar files in /lib

See here for context: https://twitter.com/bobbytank42/status/564066267645378560

jonbullock:
Wildcards isn't supported when using manifest classpath, need to set classpath in exec script as wildcards are allowed here.

ancho:
At what location is the lib/ directory expected? Relative to the root project? Or is it the lib directory of the distribution?

jonbullock:
It's the /lib directory of the distribution.

jonbullock:
@ancho are you working on this? Or did I read your comment in #381 wrong?

ancho:
Jep. I'm on it..

PR #385, opened 2017-06-16
pickup jars in distribution lib directory and add asciidoctorj-diagram as a distribution dependency

ancho:
* adds lib/* to the application classpath ( #230 )
* bumped asciidoctorj to 1.5.5
* added asciidoctorj-diagram 1.5.4 as distribution dependency

While I was working on including the lib directory, I was curious how to add library dependencies to the distribution only. So I added a new configuration called `dist` with asciidoctorj-diagram as an experiment.

That way people do not need to download the library manually but it isn't a dependency of jbake-core.

Addresses a part of #381. Not native but with asciidoctor.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,7 @@ plugins {
     id 'com.github.ben-manes.versions'  version '0.14.0'
 }
 
+
 if( JavaVersion.current().java7Compatible ) {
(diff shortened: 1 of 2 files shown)


jonbullock:
Excellent, have wanted to add this support for a while now.

Issue #281, opened 2016-04-25
Images from content folder are not copied to output folder


halyph:
I've just started using jbake and have the next issue.
It is very convenient to have adoc files and related images collocated in the same folder. E.g nice github rendering feature is supported out of the box. Also, atom editor, asciidocfx support image rendering when they are in the same folder.

But, jbake does not copy images from `content` folder to `output` folder. I've been using default settings.

Could you please clarify if this issue can be fixed via some settings? Or it is require jbake modification?

jonbullock:
To achieve this you should place the images into a folder structure underneath assets that matches what you have under content.

manikmagar:
@jonbullock I am keeping my jbake template and blog content in separate github repos intentionally. That gives me ability to change my template easily. I indeed planned to have content images in my content repo. I did not realize that images will not get copied while baking. Personally I think it should be allowed as I can see a clear benefit from it which is segregation of a template and content. Do you think this feature can be added to copy any static content from contents directory to output?

haumacher:
@manikmagar  This question seems not to be much about sparation of template and content, but enforced separation of text content and image content.

When baking the source structure is required to be 

```
project
 +-- assets
 |       +-- somefolder
 |               +-- myImage.jpg
 +-- content
         +-- somefolder
                 +-- myPageReferingMyImage.md
```
This structure is really hard to maintain, if the folder structure is deep and images are closely related to content pages. One has to duplicate the `content` folder structure under the `assets` directory. This has the additional drawback mentioned by @halyph that when browsing the repository in Github, all image references are broken, because the Github viewer expects images in the same directory as the content. 

@jonbullock  Therefore, I'm supporting @halyph's RFE to allow placing assets directly in the `contents` folder and copying everything that is not processed by JBake directly to the output folder.

manikmagar:
@haumacher I forgot to mention I consider images related to post as content, in that sense I mentioned separation. I think templates (a.k.a. themes) and contents (text as well images related to text) should be allowed to manage seperately. `assets` folder can very well be used to keep any common images related to the template. For example, If I am using a background image on my header, I am ok to keep that in the `assets` because when I change my template from A to B, that image may not be relavant to template B.

Now, if I have a post (written in md or in any format) and I have 2 images inline to that post and are post-specific then I wouldn't want to store those inside `assets` directory because when I change my template, `assets` will be a mix of template as well content-related resources and thus will require more efforts to maintain (like @haumacher, you said its difficult to maintain).

[...]

PR #390, opened 2017-06-27
Allow folders to be ignored and have static files in content folder

### LINKED ISSUE #324: Allow certain folders to be ignored from being processed either as content or assets
Allow certain folders to be ignored from being processed either as content or assets.

manikmagar:
Closes #324  and #281.

This PR adds two functionalities:

**1. Allow to ignore certain folders**:
For any folder in content or assets directory, if you add a file named `.jbakeignore` then that directory, as well as its children folders, will be skipped from copying to output. If it is a folder in the content directory, then the content files (html, ascii, md etc) will NOT be rendered. Also, any hidden files from content directory will be ignored.

**2. Allow static files (eg. images) in content directory**
All non-content files (anything other than .html, asciidocs or markdown ) will be copied to output folder as-is. File paths relative to content will be maintained in the output. For example, `/content/blog/2013/images/abc.jpg` will be copied to to `/output/blog/2013/images/abc.jpg`.


jonbullock:
I've had a quick scan over the code and it looks good to me, I'll have a proper look though as well and complete the review here ASAP.

jonbullock:
I've resolved the conflict locally but I won't merge it in until @ancho has had a chance to review the PR first.

manikmagar:
Sounds good 👍

src/main/java/org/jbake/app/Asset.java
```diff
@@ -81,6 +81,35 @@ public boolean accept(File file) {
             }
         }
     }
+    
+    public void copyAssetsFromContent(File path){
+    	copyNonContentFiles(path, destination);
+    }
+    
+    private void copyNonContentFiles(File sourceFolder, File targetFolder) {
```
ancho: This is a duplication of the copy method in Line [55](https://github.com/jbake-org/jbake/pull/390/files#diff-c0fef1d7b85aafda758d7f9855f5d31eL55). The only difference is the FileFilter applied to the sourceFolder.

So it would be nice to refactor it to something like:

```
	/**
	 * Copy all files from supplied path.
	 *
	 * @param path	The starting path
	 */
	public void copy(File path) {

	    FileFilter filer = new FileFilter() {
            @Override
            public boolean accept(File file) {
                return (!ignoreHidden || !file.isHidden()) && (file.isFile() || FileUtil.directoryOnlyIfNotIgnored(file));
            }
        };
	    copy(path, destination, filer);
    }

    public void copyAssetsFromContent(File path){

        FileFilter filer = FileUtil.getNotContentFileFilter();
        copy(path, destination, filer);
    }

    private void copy(File sourceFolder, File targetFolder, FileFilter filter) {
        final File[] assets = sourceFolder.listFiles(filter);
        if (assets != null) {
            Arrays.sort(assets);
            for (File asset : assets) {
                final File target = new File(targetFolder, asset.getName());
                if (asset.isFile()) {
                    final StringBuilder sb = new StringBuilder();
                    sb.append("Copying [").append(asset.getPath()).append("]... ");
                    try {
                        FileUtils.copyFile(asset, target);
                        sb.append("done!");
                        LOGGER.info(sb.toString());
                    } catch (IOException e) {
                        sb.append("failed!");
                        LOGGER.error(sb.toString(), e);
                        errors.add(e);
                    }
                } else if (asset.isDirectory()) {
                    copy(asset, target, filter);
                }
            }
        }
    }
```

manikmagar: Sounds good ... I will take a look and make the changes. Thanks!

manikmagar:
@ancho I refactored the code to consider your review comments, ran tests to make sure everything is still same. Can you please take a look at the new code? Thanks.

src/main/java/org/jbake/app/Asset.java
manikmagar: Made the changes and committed back to this PR. Please check now.

PR #393, opened 2017-07-31
watches templates in addition to content directories to trigger rebuidl

pblack:


jonbullock:
Thanks for taking the time to create and submit this PR, unfortunately this had already been implemented in #384 (to address #346).

PR #405, opened 2017-09-12
Fixes #389

### LINKED ISSUE #389: Paged Index affects `posts` data model on index page
When index is paginated, `published_posts` data model is paginated and provided to template for rendering. This is also affecting the `posts` data model if used on the index page. 

For example, look at he 'Recent posts' section ine left side bar on [this template](https://manikmagar.github.io/jbake-future-imperfect-template/). If you navigate through pages, you will notice that it is always same as the post returned by `published_posts` in main area. This index is paginated with 1 post on each page.

jonbullock:
Added additional methods to allow control over when the paging filter is applied. Removed paging on all_content as I believe it was a mistake being added to this as paging is currently based around 'post' type only. Kept the rest of the method signatures the same though.

### DIFF
--- a/src/main/java/org/jbake/app/ContentStore.java
+++ b/src/main/java/org/jbake/app/ContentStore.java
@@ -134,13 +134,16 @@ public class ContentStore {
 
     public DocumentList getDocumentStatus(String docType, String uri) {
         return query("select sha1,rendered from " + docType + " where sourceuri=?", uri);
-
     }
 
     public DocumentList getPublishedPosts() {
         return getPublishedContent("post");
     }
-
+    
+    public DocumentList getPublishedPosts(boolean applyPaging) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jbake/app/ContentStore.java
```diff
@@ -159,17 +162,29 @@ public DocumentList getPublishedPages() {
     }
 
     public DocumentList getPublishedContent(String docType) {
+        return getPublishedContent(docType, false);
+    }
+    
+    public DocumentList getPublishedContent(String docType, boolean applyPaging) {
         String query = "select * from " + docType + " where status='published' order by date desc";
```
manikmagar: In future, when paging for multiple content types is supported, LIMIT and SKIP variables can be made docType specific. As with this release, only paging of POST is supported, this fix should do the work!

Issue #408, opened 2017-09-29
Baking process is not possible with Java 9


tuxtor:
I'm using a fresh JBake (2.5.1 SDKMan) installation over Fedora 26 and the newly released Oracle JDK 9.

So far when I use JDK 9 to trigger baking process I'm receiving this as output:

> $ jbake -b 
> JBake v2.5.1 (2017-01-31 23:24:52PM) [http://jbake.org]
> 
> 19:37:56.676 WARN  c.o.o.c.d.r.ODatabaseRecordAbstract$2 - Current implementation of storage does not support sbtree collections
> An unexpected error occurred: Cannot create database

However If i switch to JDK 8 (Oracle and OpenJDK) all works properly.

ancho:
Hello @tuxtor. We are switching to a newer version of orient db. See #337. The pull request is quite old. The newest version in 2.x is 2.2.28.
There is an open issue at orientdb https://github.com/orientechnologies/orientdb/issues/7780 addressing problems with jdk 9.

I made a quick test with 2.2.28. There are several warnings from orientdb but baking a site was successful. So we need to be patient and wait for orientdb to fix the jdk9 issues.

```
JBake v2.6.0-SNAPSHOT (2017-10-01 10:29:28vorm.) [http://jbake.org]

10:45:31.581 WARN  c.o.common.util.OMemory - Unable to determine the amount of installed RAM.
java.lang.reflect.InaccessibleObjectException: Unable to make public long com.sun.management.internal.OperatingSystemImpl.getTotalPhysicalMemorySize() accessible: module jdk.management does not "opens com.sun.management.internal" to unnamed module @70f02c32
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:337)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:281)
	at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:198)
	at java.base/java.lang.reflect.Method.setAccessible(Method.java:192)
	at com.orientechnologies.common.util.OMemory.getPhysicalMemorySize(OMemory.java:65)
	at com.orientechnologies.orient.core.engine.OMemoryAndLocalPaginatedEnginesInitializer.configureDefaultDiskCacheSize(OMemoryAndLocalPaginatedEnginesInitializer.java:79)
[...]

PR #409, opened 2017-10-01
update dependencies. switch to junit5 and jetty9

ancho:
While testing jdk9 support (#408) I updated all dependencies and switched to jetty9.


ancho:
Ok. Forget it. I'm obviously not awake. Totally forgot we are in java 7 support land here ¯\_(ツ)_/¯

PR #410, opened 2017-10-02
update versions

ancho:
Another try to update all dependencies to jdk 7 compatible versions.


jonbullock:
Thanks for going through all the pain to get this all running on JDK 7 👍

Issue #428, opened 2018-03-04
configure continous code checks with sonarcloud


ancho:
See https://docs.travis-ci.com/user/sonarcloud/

jonbullock:
See #449

PR #430, opened 2018-03-06
run sonar scanner and report with travis. enable pull-request comments

ancho:
This is an attempt to address #428 
Configured gradle build and travis. Fixed some sonar issues to see what happens.
We are using the stage beta feature now. Added a Quality check stage which runs the sonarqube task
if Testing stage was successful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,56 @@
 language: java
 
-os: linux
 dist: trusty
 
 addons:
   hosts:
-    - jbake-builder
+  - jbake-builder
   hostname: jbake-builder
-
-matrix:
-    include:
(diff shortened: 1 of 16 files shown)


ancho:
There is a restriction from travis, that prevents pull request comments from sonar github-plugin.

> Please note that encrypted environment variables are not available for pull requests from forks.
See https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions.

It is only possible with internal pull-requests. That way it's also not possible to send analysis information to sonarcloud from a pull-request.

But it works from master. Tested it with my fork and an internal pull-request.

PR #431, opened 2018-03-08
Fix/407 jvm locale in properties

### LINKED ISSUE #407: Allow JVM locale to be set via jbake.properties
See #182

rustamzh:
Attempt to close #407. I have compared it with specifying JAVA_OPTS inside gradlew - no difference in the test.  Welcome any reviews.

### DIFF
--- a/jbake-core/src/main/java/org/jbake/launcher/Main.java
+++ b/jbake-core/src/main/java/org/jbake/launcher/Main.java
@@ -12,10 +12,7 @@ import org.slf4j.bridge.SLF4JBridgeHandler;
 
 import java.io.File;
 import java.io.StringWriter;
-import java.text.MessageFormat;
-import java.util.List;
 import java.util.Locale;
-import java.util.NoSuchElementException;
 
 /**
  * Launcher for JBake.
--- a/jbake-core/src/main/java/org/jbake/launcher/Main.java
+++ b/jbake-core/src/main/java/org/jbake/launcher/Main.java
@@ -2,6 +2,7 @@ package org.jbake.launcher;
 
 import org.apache.commons.configuration.CompositeConfiguration;
 import org.apache.commons.configuration.ConfigurationException;
+import org.apache.commons.lang.LocaleUtils;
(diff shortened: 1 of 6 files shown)


jbake-core/src/main/java/org/jbake/launcher/Main.java
```diff
@@ -177,4 +176,12 @@ private void initStructure(CompositeConfiguration config, String type, String so
 		}
 	}
 
+	private void setupLocale(CompositeConfiguration config){
+		if (config.containsKey("jvm.language") && config.containsKey("jvm.country")){
+			String language = config.getString("jvm.language");
```
jonbullock: Could you convert these config strings to constants like here: https://github.com/jbake-org/jbake/blob/master/jbake-core/src/main/java/org/jbake/app/ConfigUtil.java#L28

jbake-core/src/main/java/org/jbake/launcher/Main.java
```diff
@@ -177,4 +176,12 @@ private void initStructure(CompositeConfiguration config, String type, String so
...
+			String country = config.getString("jvm.country");
+			Locale newLocale = new Locale(language, country);
```
jonbullock: You do have access to [LocaleUtils](https://commons.apache.org/proper/commons-lang/javadocs/api-3.7/org/apache/commons/lang3/LocaleUtils.html) and can be used like this if you want: https://github.com/jbake-org/jbake/blob/master/jbake-core/src/main/java/org/jbake/template/ThymeleafTemplateEngine.java#L63

rustamzh:
Thank you for the review. I have revised implementation.

ancho:
Hi. Thanks. Maybe we should setup the locale in the Oven bake method? The gradle plugin does not make use of the Main class.

rustamzh:
Hello. Moved locale setup to the Oven.

rustamzh:
Sorry for closing PR. Thought it will close commenting box.

PR #433, opened 2018-03-12
Fix travis jdk9 build -  java.lang.OutOfMemoryError: Direct buffer memory

ancho:
A jdk9 travis build failed with an OutOfMemoryError.
https://travis-ci.org/jbake-org/jbake/jobs/352334663#L1403

Rearanged test setup to start only one orientdb per test class.
Adjusted Heapspace and MaxDirectMemorySize to fit travis environment.
Updated orientdb to 2.2.33 which include some fixes related to memory estimation in container environments. https://github.com/orientechnologies/orientdb/issues/7812

### DIFF
--- a/jbake-core/src/main/java/org/jbake/app/ContentStore.java
+++ b/jbake-core/src/main/java/org/jbake/app/ContentStore.java
@@ -40,7 +40,11 @@ import org.jbake.model.DocumentTypes;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.*;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * @author jdlee
@@ -54,7 +58,7 @@ public class ContentStore {
 
     public ContentStore(final String type, String name) {
         startupIfEnginesAreMissing();
-        OPartitionedDatabasePool pool = new OPartitionedDatabasePoolFactory().get(type+":"+name, "admin", "admin");
+        OPartitionedDatabasePool pool = new OPartitionedDatabasePoolFactory().get(type + ":" + name, "admin", "admin");
         pool.setAutoCreate(true);
         db = pool.acquire();
 
@@ -84,6 +88,10 @@ public class ContentStore {
     }
 
     public final void updateSchema() {
+        if (db.isClosed()) {
+            db.create();
+        }
+
         OSchema schema = db.getMetadata().getSchema();
 
         for (String docType : DocumentTypes.getDocumentTypes()) {
@@ -97,6 +105,7 @@ public class ContentStore {
     }
 
     public void close() {
+        db.activateOnCurrentThread();
         db.close();
         DBUtil.closeDataStore();
     }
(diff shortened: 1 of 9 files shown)


jbake-core/src/main/java/org/jbake/app/ContentStore.java
```diff
@@ -97,6 +105,7 @@ public final void updateSchema() {
     }
 
     public void close() {
+        db.activateOnCurrentThread();
```
manikmagar: Please change this to call private method `activateOnCurrentThread()` to have a single entry point.

jbake-core/src/main/java/org/jbake/app/ContentStore.java
```diff
@@ -113,17 +122,23 @@ private void startupIfEnginesAreMissing() {
...
         // We need to startup Orient again.
-        if ( Orient.instance().getEngines().size() == 0 ) {
+        if (Orient.instance().getEngines().isEmpty()) {
             Orient.instance().startup();
         }
         OLogManager.instance().setWarnEnabled(true);
     }
 
     public void drop() {
+        db.activateOnCurrentThread();
```
manikmagar: Please change this to call private method `activateOnCurrentThread()` to have a single entry point.

jbake-core/src/main/java/org/jbake/app/ContentStore.java
```diff
@@ -221,22 +236,25 @@ public void insertSignature(String currentTemplatesSignature) {
...
 
     private void executeCommand(String query, Object... args) {
+        activateOnCurrentThread();
         db.command(new OCommandSQL(query)).execute(args);
     }
 
     public Set<String> getTags() {
         DocumentList docs = this.getAllTagsFromPublishedPosts();
-        Set<String> result = new HashSet<String>();
+        Set<String> result = new HashSet<>();
```
manikmagar: Is this an intentional change? I guess not.

jbake-core/src/main/java/org/jbake/app/ContentStore.java
```diff
@@ -245,7 +263,7 @@ private void executeCommand(String query, Object... args) {
     }
 
     public Set<String> getAllTags() {
-        Set<String> result = new HashSet<String>();
```
manikmagar: Is this an intentional change? I guess not.

jbake-core/src/test/java/org/jbake/app/ContentStoreTest.java
```diff
@@ -2,25 +2,36 @@
...
 import org.jbake.FakeDocumentBuilder;
 import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 
 public class ContentStoreTest {
```
manikmagar: General Thought: Should we create a Parent Test class and move ContentStore related changes to parent and thus do not repeat those in all tests? Or it is a good idea to keep tests independent?

jbake-core/src/main/java/org/jbake/app/ContentStore.java
ancho: This is intentional. Java 7 introduced "Type Inference" and the Diamond operator.
https://docs.oracle.com/javase/tutorial/java/generics/genTypeInference.html

It's also a code smell at sonarqube:

```
Java 7 introduced the diamond operator (<>) to reduce the verbosity of generics code. For instance, instead of having to declare a List's type in both its declaration and its constructor, you can now simplify the constructor declaration with <>, and the compiler will infer the type.
```

jbake-core/src/main/java/org/jbake/app/ContentStore.java
ancho: See previous answer.

jbake-core/src/test/java/org/jbake/app/ContentStoreTest.java
ancho: It is generally a good idea to keep tests as independent as possible. But maybe you are right. Will consider it.

Issue #437, opened 2018-03-22
All jbake-core dependencies are forced on you when used in another Maven project


jonbullock:
Dependencies that provide content format support and template engine support were optional prior to 2.6.0 but are now forced on you when you include jbake-core in a project (as jbake-maven-plugin does) so you can't include just the dependencies that you require (i.e. include Freemarker but not Groovy).

Had a quick search and came across this: https://github.com/gradle/gradle/issues/867

manikmagar:
Could that be the reason for binary file size difference? 2.6.0 download size is 69MB, compared to 2.5.1 having 40MB only.

ancho:
ah. honestly I wasn't aware of the optional maven dependencies while switching to gradle. so the nebula plugin seems the way to go.

as for the size difference. we had a few dependency changes since 2.5.1. asciidoctorj-diagram and felxmark (as replacement for pegdown).

See directory and file sizes for [2.5.1](https://paste.rs/3gZ) and [2.6.0](https://paste.rs/Joa) and the [dependency graph](https://paste.rs/CcO)

We could replace `flexmark-all` with a minimal set of dependencies we really need. We don't need all extensions and converters.

jonbullock:
I'd completely forgotten about it... only noticed as I'm updating my example projects that use the plugins for a talk and forgot to include the other dependencies and thought to myself how is it even baking a site without them being declared.

Issue #224, opened 2015-08-03
Move to Java 8 for v2.5.0 or v3.0.0


msgilligan:
By its nature JBake is a standalone tool installed on an individual machine that generates static output that is pushed to the server. So there's little worry about runtime requirements other than in the build tools themselves. Java 7 is reaching end-of-maintenance (or whatever Oracle calls it) and Java 8 offers significant improvements.

I see no reason not to move to Java 8 for v2.5.0, but at the very least v3.0.0 should use it (and skip Java 7)

This issue started out as a comment on Issue #107, but I thought it should be a new, separate issue. Java 8 has been adopted very rapidly in the 16 months since that issue was created and we should consider moving to it as quickly as possible.

jangalinski:
+1

jonbullock:
Absolutely fine with me, I'll give the community chance to voice any concerns on the other thread before closing it in favour of this.

melix:
Just to make it clear : as soon as you require 1.8, you force all
downstream apps that integrate JBake to upgrade. It means the Gradle plugin
as well as the maven one. It's not necessarily a good thing.

Your call anyway.
Le 16 août 2015 21:08, "Jonathan Bullock" <notifications@github.com> a
écrit :

haumacher:
For JBake users, there is no benefit when increasing the minimum Java compatibility level. Currently built against Java 6, JBake should be runnable on any Java 6, Java 7 and Java 8 VM (if that's not the case it's an issue that should be fixed). 

When compiling against Java 8, you require a Java 8 VM for runtime. The only benefit upgrading the JVM version is the possibility to uses Java 7 or Java 8 language features for development, e.g. try with resource, simplified type parametrization or lambda expressions. While this might slightly simplify the code here and there, I think it's not worth breaking the compatibility to embedding apps that cannot upgrade to Java 8 in the near future.

msgilligan:
>  the possibility to use Java 7 or Java 8 language features for development

This is a huge advantage to the project IMO.

> compatibility to embedding apps

Aside from the Gradle and Maven plugins (which should be straightforward to upgrade) what embedding apps are there? @haumacher Do you have a specific embedding app in mind?

My impression is that JBake is used on individual machines to generate static sites and is not embedded in complex server applications which are slow to upgrade. It would be great to see JBake have the cleanest, most modern code base possible and to favor rapid development to better compete with the Node based ecosystem and to attract more contributors and more eyeballs on the code.

[...]

PR #438, opened 2018-03-23
reduce distribution size

ancho:
As noted in #437 the size of the distribution package increased.
Modified dependencies of jbake-core to get it smaller.

* replaced orientdb-graphdb with orientdb-core
  we don't use the graphs and do not need the server jar
* replaced flexmark-all with core and pegdown profile
  we don't need converters and stuff like that
* excluded transitive junit dependency
  make no sense at runtime
* updated
  * asciidoctorj-diagram 1.5.4.1 -> 1.5.6
  * flexmark 0.28.38 -> 0.32.18
  * groovy 2.4.13 -> 2.4.14
  * jsoup 1.10.3 -> 1.11.2
  * mockito 2.13.0 -> 2.16.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
 // common variables
 ext {
     asciidoctorjVersion = '1.5.6'
-    asciidoctorjDiagramVersion = '1.5.4.1'
+    asciidoctorjDiagramVersion = '1.5.8'
     commonsIoVersion = '2.6'
     commonsConfigurationVersion = '1.10'
     commonsLangVersion = '3.7'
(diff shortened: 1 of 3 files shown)


ottlinger:
+1 

AssertJ-core could be updated as well .... https://joel-costigliola.github.io/assertj/assertj-core-quick-start.html
or does the project have to stick with JDK7?

manikmagar:
👍 

@ottlinger With 3.0.0, plan is to upgrade to full Java 8 (see #224). We can upgrade assertj along with that.

ottlinger:
Would it be possible to push out a release with these changes? 2.6.1? Thanks.

PR #443, opened 2018-04-04
Fixes #437

### LINKED ISSUE #437: All jbake-core dependencies are forced on you when used in another Maven project

jonbullock:
Fixes #437 

Implemented Nebula plugin to enable support for optional flag on certain dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ plugins {
     id "com.jfrog.bintray"              version "1.8.0"     apply false
     id "com.github.kt3k.coveralls"      version "2.8.2"     apply false
     id 'com.github.ben-manes.versions'  version '0.17.0'
-    id "nebula.optional-base"           version "3.3.0"
+    id "nebula.optional-base"           version "3.3.0"     apply false
 }
 
 // common variables
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,7 @@ plugins {
     id "com.jfrog.bintray"              version "1.8.0"     apply false
     id "com.github.kt3k.coveralls"      version "2.8.2"     apply false
     id 'com.github.ben-manes.versions'  version '0.17.0'
+    id "nebula.optional-base"           version "3.3.0"
 }
 
 // common variables
@@ -60,6 +61,7 @@ allprojects {
 subprojects {
 
     apply plugin: 'java'
+    apply plugin: 'nebula.optional-base'
 
     if( JavaVersion.current().java7Compatible ) {
         apply from: "$rootDir/gradle/signing.gradle"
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -5,6 +5,7 @@ plugins {
     id "com.jfrog.bintray"              version "1.8.0"     apply false
     id "com.github.kt3k.coveralls"      version "2.8.2"     apply false
     id 'com.github.ben-manes.versions'  version '0.17.0'
+    id "nebula.optional-base"           version "3.3.0"
```
ancho: If you add a plugin here it gets applied. Our base project is not a java project and does not need the plugin.
`apply false` prevents that.

build.gradle
```diff
@@ -77,8 +79,8 @@ subprojects {
     dependencies {
         compile "org.slf4j:slf4j-api:$slf4jVersion"
         compile "org.slf4j:jul-to-slf4j:$slf4jVersion"
-        compile "ch.qos.logback:logback-classic:$logbackVersion"
-        compile "ch.qos.logback:logback-core:$logbackVersion"
+        compile "ch.qos.logback:logback-classic:$logbackVersion", optional
+        compile "ch.qos.logback:logback-core:$logbackVersion", optional
```
ancho: Why is logback optional?

build.gradle
jonbullock: Ah OK I'll add `apply false` to the plugin.

build.gradle
jonbullock: I had to double check this as it surprised me when I saw it was set as optional in 2.5.1, so the Logback dependency isn't forced on the 3rd party using jbake-core. The 3rd party is free to use any framework they like.

jonbullock:
Essentially yes the 3rd party needs to know what dependencies to use for the specific features (i.e. AsciidoctorJ for AsciiDoc support)

ancho:
hmm. is there an easy way to add all optionals or get a list of those? i could imagine it's slightly annoying going into a trial and error loop because you forget that one dependency you are not aware of and everything breaks.

jonbullock:
I'm not aware of any way to list the optional dependencies.... but there may be.

This was partially addressed by the documentation: https://jbake.org/docs/2.6.0/#maven_coordinates

jonbullock:
Yeah I noticed that section.

I'll make sure the download page, under the Maven section, has a link to the specific section of the docs that details the optional dependencies.

PR #450, opened 2018-04-15
Fix #365 issue: add default 404 file generation

### LINKED ISSUE #365: Generate 404 page not found html
No matter where static website/blog is hosted, you would always want a page to display when 404 page not found error occurs. 

Github pages may be the most popular ( & free :) ) place to host static sites. Github does support adding a custom 404 page not found file in github-pages - See [this](https://help.github.com/articles/creating-a-custom-404-page-for-your-github-pages-site/). It is simple, 404.html should exist at the root of your gh-pages.

There are few ways to do this with -

1. Add a content page at the root of content folder and let it render with page template. Example [here](https://github.com/manikmagar/jbake-future-imperfect-template/blob/master/content/404.html). 
2. Register a new custom type 404, create a template file for the custom type and add a content 404 file. This will render 404.html with the dedicated template for 404.
3. Add a support in JBake to by default generate a 404.html file at the root. Templates can have a 404.ftl (or similar template) which is rendered by JBake. No need to add any content. 

I feel JBake should by default generate 404 (option 3), that way it feels to be part of framework.

Any thoughts?

rustamzh:
This PR tries to close #365. This PR stands good on tests against core functionality, but fails on dist test. I assume that happens because there is no standard 404 template in example repositories. I have quickly written 404 templates for all template engines that are present in /fixture folder in test package.

### DIFF
--- a/jbake-core/src/main/java/org/jbake/app/ConfigUtil.java
+++ b/jbake-core/src/main/java/org/jbake/app/ConfigUtil.java
@@ -102,6 +102,10 @@ public class ConfigUtil {
 		 */
 		String INDEX_FILE = "index.file";
 		/**
+		 * Output filename for error 404 file, is only used when {@link #RENDER_ERROR404} is true
+		 */
+		String ERROR404_FILE = "error404.file";
+		/**
 		 * File extension to be used for all output files
 		 */
 		String OUTPUT_EXTENSION = "output.extension";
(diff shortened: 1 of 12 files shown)


manikmagar:
Thank you @rustamzh for taking time to implement this. I see some tests have failed, would you mind running `./gradlew check` in local and fix those test cases?

manikmagar:
Thank you @rustamzh.

manikmagar:
Changes looks good to me. @jonbullock Can you take a look and see if we can include this in 2.7.0?

jonbullock:
Yeah looks OK to me.

Issue #455, opened 2018-04-17
Problem with static content PR #390


gilbertoca:
Hello guys,

First time running version 2.6.1 . The image files are copied, but the behavior is not the expected.
In the md file I have
```md
[<img src="CompileOnSave.png"  title="CompileOnSave" width="804" height="563" class="alignright size-full wp-image-701" />
```
But the final output has
```xml
<a href="CompileOnSave.png"><img src="http://localhost:8820/CompileOnSave.png" title="CompileOnSave" width="804" height="563" class="alignright size-full wp-image-701">
```
See the URL (without the /2011/05/ folder).
The image is not rendered a priory. But, to my surprise,  hover the mouse on it shows the right URL result
```xml
http://localhost:8820/2011/05/CompileOnSave.png
```
![screenshot_20180417_150013](https://user-images.githubusercontent.com/3083703/38887828-47f61a12-4250-11e8-97b8-deb60fc57bc5.png)

manikmagar:
Hi Gilberto,

Could you try changing URL to "./CompileOnSave.png"?

See some examples here - https://jbake.org/docs/2.6.1/#images_path

Regards,
Manik Magar
https://manik.magar.me

gilbertoca:
@jonbullock @manikmagar I think this approach is contra-intuitive, I have tons of images in my site. See this [reference on file path](https://www.w3schools.com/html/html_filepaths.asp) as well .

msgilligan:
This seems related to my Issue https://github.com/jbake-org/jbake/issues/440

manikmagar:
Oh ya, I agree that path with and without `./` means the same. I will take a look at that.

ancho:
Oh. Another problem is the concatenation of `site.host` to the image URL. That leads to a problem serving the site with `jbake -s` as the image may not be present on that site when it is new.

manikmagar:
`site.host` was appended to resolve the issues like when an article is served from other programs such as RSS readers or article is also served from other location within the site. Now, since the images have FQ url, it doesn't matter from where article is accessed. 

@ancho I think `jbake -s` ignores the site.host value from properties and uses `localhost`. There shouldn't be that problem as long as images are in the current source because locally baked and served site will have `localhost` in the URL.

[...]

PR #459, opened 2018-04-24
Fixes Image Sources

### LINKED ISSUE #440: Option to generate relative image links in output
I want to generate an output site file tree that has _relative_ links to the images. I don't want to use server mode -- I want an _accurate_ preview loading from `file://` URLs. I also want the option to post to a staging and production server (using Gitlab CI) so the image links shouldn't be hardcoded to the production server.

But the images are pulled from the previously published version of the site running on `site.host`.

According to the documentation `site.host` should contain the URL of the server and is overridden for server mode. But there appears to be no (easy) way to request the generation of relative image links in the output.

I'm in the process of updating to JBake 2.6.0.

Note https://github.com/jbake-org/jbake-gradle-plugin/issues/32 -- Gradle plugin doesn't support server mode.

manikmagar:
Fixes #440 and #455 

1. Do not modify image source if it starts with `http://` or `https://`.
2. Prepend Resolved root path and page URL to image source, unless it starts with `/`. Eg. `CompileOnSave.png` -> `../../2011/05/CompileOnSave.png`,  `./CompileOnSave.png` -> `../../2011/05/CompileOnSave.png` but `/CompileOnSave.png` -> `/CompileOnSave.png`
3. `img.path.prepend.host` (default `true`, otherwise it will not be backward compatible) decides if `site.host` should be prepended to image source with Point 2 above. When set to false, `site.host` will not be appended. [this along with addition root path should solve #440 ]


msgilligan:
What is "Resolved root path"?

jonbullock:
I'm guessing @manikmagar is referring to ```content.rootpath``` which is the path to the [root of the project](https://jbake.org/docs/2.6.1/#global)

manikmagar:
Thanks @jonbullock. Yes I meant `content.rootpath`.

jbake-core/src/test/java/org/jbake/util/HtmlUtilTest.java
```diff
@@ -40,34 +40,50 @@ public void shouldNotAddBodyHTMLElement(){
...
+        fileContent.put(Attributes.ROOTPATH, "../../../");
+        fileContent.put(Attributes.URI, "blog/2017/05/first_post.html");
+        fileContent.put(Attributes.BODY, "<div> Test <img src='./first.jpg' /></div>");
+        config.setProperty(ConfigUtil.Keys.IMG_PATH_PREPEND_HOST,false);
+
+        HtmlUtil.fixImageSourceUrls(fileContent, config);
+
+        String body = fileContent.get(Attributes.BODY).toString();
+
+        assertThat(body).contains("src=\"../../../blog/2017/05/first.jpg\"");
```
ancho: This will only work for the first_post.html file. if it is an post shown with content on the index page the url would be wrong openening the file directly from a browser. on the index page it should be `blog/2017/05/first.jpg`.

jonbullock:
Merged in manually as part of #535

PR #461, opened 2018-04-30
Update dependency in README.asciidoc

srbala:
Update dependency of markdown in README.asciidoc


manikmagar:
:+1:  Thanks for the update.

PR #478, opened 2018-07-22
Document minimal Gradle version.

OndraZizka:
4.9 is the current. Maybe the whole 4.x works  but 3.x does not.


ancho:
Yes. I noticed this while working on #471. The pr includes an update to gradle 4.9.

PR #487, opened 2018-07-23
Add default.type, commented out, for documentary purposes

OndraZizka:
This file seems to be the authoritative source of configuration options list.
Therefore maybe it should contain all options, even if commented out, so that the user who sees this file knows about their existence.

I also separated some logical groups with a newline, IMO it improves readability a bit.


ancho:
Thank you very much. If you have some time left have a look at #341. Let me know what you think.
I want to cleanup more things like that.

jonbullock:
Thanks for raising this 👍

OndraZizka:
Maybe we could shuffle the config a bit to have markups together, and other logical groups. And maybe have an empty line before each item.

jonbullock:
Yeah good idea @OndraZizka I'll create a separate issue for that.

PR #513, opened 2018-08-01
Print stacktrace on "unexpected error", too (in which case it is actu…

OndraZizka:
…ally important)

The handling of errors will have to improve in general. printStackTrace is basically telling the user "go fetch code and fix it :)"


jonbullock:
Yeah the error reporting does need improved.

PR #518, opened 2018-08-01
Fix #499 file names encoding.

### LINKED ISSUE #499: Bug: File names are not URL encoded in links.
This causes that  `content/foo#bar.md`, which is a valid filename, and thus can be served as a static content, ends up being linked to as `.../foo#bar.html`, which is like `.../foo` and won't work.

Fix: (I'll do a PR)
```
    private String createUri(String uri) {
        ///return uri.substring(0, uri.lastIndexOf('.')) + config.getString(Keys.OUTPUT_EXTENSION);
        try {
            return "/" + FilenameUtils.getPath(uri)
                    + URLEncoder.encode(FilenameUtils.getBaseName(uri), StandardCharsets.UTF_8.name())
                    + config.getString(Keys.OUTPUT_EXTENSION);
        }
        catch (UnsupportedEncodingException e) {
            throw new RuntimeException("Missing UTF-8 encoding??", e); // Won't happen unless JDK is broken.
        }
    }
```

OndraZizka:
Fixes https://github.com/jbake-org/jbake/issues/499

### DIFF
--- a/jbake-core/src/main/java/org/jbake/app/Crawler.java
+++ b/jbake-core/src/main/java/org/jbake/app/Crawler.java
@@ -1,6 +1,9 @@
 package org.jbake.app;
 
 import java.io.File;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


jonbullock:
👍

PR #521, opened 2018-08-01
Add DebugUtil with map printing, since JBake code uses maps heavily

OndraZizka:
I add and remove calls to this when debugging.


jonbullock:
👍

PR #522, opened 2018-08-01
Normalize whitespace - replace tabs with 4 spaces.

OndraZizka:
As per discussion at jbake-dev.
This only replaces the tabs, doesn't reformat.


OndraZizka:
Replaced by https://github.com/jbake-org/jbake/pull/555 .

Issue #489, opened 2018-07-24
default.status=published doesn't work


OndraZizka:
I have `default.status=published` in `jbake.properties`.

However, if a document does not contain `status=published`, it is not rendered.

Also, the logic of the error seems to be wrong. When I remove the `status=published` header line, the error actually occurs during the Freemarker rendering:
```
FTL stack trace ("~" means nesting-related):
        - Failed at: ${content.title}  [in template "page.ftl" at line 6, column 21]
```
This is the document it fails with:
```
title=Test Post Markdown, no status
date=2018-07-21
type=post
tags=blog,test
@@@

This is a test post in Markdown. For some info, see [About](about.html).
```

OndraZizka:
I think the bug is in `hasHeader()` :

            return headerValid && statusFound && typeFound;

The status is not there, so it returns false, so this doesn't happen:

        if (hasHeader) {
            // read header from file
            processHeader(config, fileContents, content);
        }

OndraZizka:
I would suggest to reduce the logic in `hasHeader()` and perhaps remove it entirely.
It duplicates `processHeader()`. It tries to validate the header but not canonically.

How about just having processHeader and then implementing a validation rather than parsing the file twice?

OndraZizka:
I'm working on a fix.

jonbullock:
Thanks for raising this bug. The idea of a validation method sounds good to me, the complication is AsciiDoc where the metadata header is not required and the metadata is embedded in the AsciiDoc content itself.

OndraZizka:
That should be okay. The validation would happen after all input is takein into account:
1. The defaults
2. Document's metadata (including AsciiDoc)
3. Any other sources of metadata - Git, location, ...
3. JBake headers block

Issue #488, opened 2018-07-23
OrientDB threw IllegalArgumentException: Cluster name is empty


OndraZizka:
I ran JBake 2.6.1 on a directory with previously saved cache.
And got the following exception.

I'm capturing it in case someone knows what that could be.
And I suggest (and offer) to handle the exception gracefully (i.e. catch at some point in JBake).

```
00:24:49.693 INFO  org.jbake.app.Crawler - Processing [/home/ondra/uw/oz.cz/jbake-based/content/blog/2018/test.texy]...  : same 
00:24:49.837 ERROR unknown.jul.logger - Exception `09BD0FA6` in storage `plocal:cache`: 2.2.33 (build 77584cd6827f647cf4aa231cf27bd6f10bc04e2c, branch 2.2.x)
java.lang.IllegalArgumentException: Cluster name is empty
        at com.orientechnologies.orient.core.storage.impl.local.OAbstractPaginatedStorage.getClusterIdByName(OAbstractPaginatedStorage.java:1696)
        at com.orientechnologies.orient.core.db.document.ODatabaseDocumentTx.getClusterIdByName(ODatabaseDocumentTx.java:1340)
        at com.orientechnologies.orient.core.metadata.schema.OSchemaShared.createClusters(OSchemaShared.java:1228)
        at com.orientechnologies.orient.core.metadata.schema.OSchemaShared.doCreateClass(OSchemaShared.java:992)
        at com.orientechnologies.orient.core.metadata.schema.OSchemaShared.getOrCreateClass(OSchemaShared.java:299)
        at com.orientechnologies.orient.core.metadata.schema.OSchemaShared.getOrCreateClass(OSchemaShared.java:269)
        at com.orientechnologies.orient.core.metadata.schema.OSchemaProxy.getOrCreateClass(OSchemaProxy.java:82)
        at com.orientechnologies.orient.core.metadata.schema.OSchemaProxy.getOrCreateClass(OSchemaProxy.java:71)
        at com.orientechnologies.orient.core.record.impl.ODocument.setClassName(ODocument.java:2009)
        at com.orientechnologies.orient.core.record.impl.ODocument.<init>(ODocument.java:161)
        at org.jbake.app.Crawler.crawlSourceFile(Crawler.java:209)
        at org.jbake.app.Crawler.crawl(Crawler.java:113)
        at org.jbake.app.Crawler.crawl(Crawler.java:118)
        at org.jbake.app.Oven.bake(Oven.java:142)
[...]

PR #530, opened 2018-08-15
[WIP] fix: hasHeader respects default.status and default.type

### LINKED ISSUE #481: WARNING printed despite default.status=published is set
I have `default.status=published` in `jbake.properties`.

I omitted the status from one file, and it is rendered, but still, jbake prints this warning:

> 05:55:13.728 WARN  org.jbake.app.Crawler - /home/ondra/uw/oz.cz/jbake-based/content/about.md has an invalid header, it has been ignored!

I think this should be muted if there's the default.

ancho:
Teaching hasHeader to respect default.status and default.type as proposed as a fix in #489.
Does not handle title validation or in general other required metadata.

AsciidoctorEngine does not need the title to be set in the header metadata. It derives the title from it's own document header parser.

So maybe we should implement the required metadata validation in the concrete MarkupEngine implementation.

Should also fix #481 and #488


jonbullock:
Haven't forgotten about the review of this, will try and get to it tomorrow.

PR #532, opened 2018-08-19
introduce DocumentModel and TemplateModel

ancho:
A little refactoring to introduce a kind of a model abstraction.

A DocumentModel is basically a HashMap with content specific content collected during crawling with explicit getters and setters.

A TemplateModel is a Hashmap with content needed during the rendering phase. It contains a DocumentModel as content.

From there on, we can extend the TemplateModel to add different kind of helper classes for use in different template engines.

PR #533, opened 2018-08-20
replace document type schema classes with one class for all documents

ancho:
Based on my previous refactoring #532 I wondered if the crawling process coudn't be much cleaner if we store all documents in the same schema class in orientdb.

So instead adding and managing a schema per document type I just changed contentstore to use one `Documents` class to store...well...all the documents we crawl.

The only breaking change here is the fact that people using `db.store=plocal` need to call `jbake -b --reset` once.

What do you think? See https://github.com/jbake-org/jbake/commit/92c25a368a54253ce728262c151bef3bcc541290 to get a better overview.


jonbullock:
Same as #532, 2.7.0 ?

ancho:
Yes. But maybe we should think about the consequences for using only one document schema.
The code may seem cleaner, but I'm not quite sure wether it's the best practice for orientdb.

I need to reread the orientdb documentation.

If we use the document model right, we should get more information out of it for template rendering purposes.
Like which page is a sibbling of another? Where is the page locatet? And things like that. Directly from the database. Which may help us to address several path issues.

PR #537, opened 2018-09-28
Wraps image path update in config option

jonbullock:
Relates to #535 and #459

### DIFF
--- a/jbake-core/src/main/java/org/jbake/app/configuration/JBakeConfiguration.java
+++ b/jbake-core/src/main/java/org/jbake/app/configuration/JBakeConfiguration.java
@@ -283,12 +283,14 @@ public interface JBakeConfiguration {
     boolean getUriWithoutExtension();
 
     /**
-     * @return Flag indicating if image paths should be prepended with {@link #getSiteHost()} value - see {@link #getImgPathUpdate()}
+     * @return Flag indicating if image paths should be prepended with {@link #getSiteHost()} value - only has an effect if
+     * {@link #getImgPathUpdate()} is set to true
      */
     boolean getImgPathPrependHost();
 
     /**
-     * @return Flag indicating if image paths should be updated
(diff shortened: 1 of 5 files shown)


jbake-core/src/main/java/org/jbake/app/configuration/JBakeConfiguration.java
```diff
@@ -283,10 +283,15 @@
     boolean getUriWithoutExtension();
 
     /**
-     * @return Flag indicating if image paths should be prepended with {@link #getSiteHost()} value
+     * @return Flag indicating if image paths should be prepended with {@link #getSiteHost()} value - see {@link #getImgPathUpdate()}
```
kwin: For me this description is not clear enough. What different effects do I get with the different combinations of boolean values for `img.path.prepend.host` and `img.path.update`?

kwin: I don't see `context.rootpath` in https://github.com/jbake-org/jbake/blob/a108a6fb594fac567cbba3a3fd870ca9b1627c08/jbake-core/src/main/java/org/jbake/app/configuration/JBakeProperty.java nor an according getter in https://github.com/jbake-org/jbake/blob/a108a6fb594fac567cbba3a3fd870ca9b1627c08/jbake-core/src/main/java/org/jbake/app/configuration/JBakeConfiguration.java. But if that is the case, I think it is worth mentioning both properties in the javadoc (i.e. site.host and context.rootpath)

jonbullock: The `rootpath` variable is calculated for each source content file, which is why it isn't in the config classes, however that variable isn't used in this implementation anymore. 

I'll update the javadoc though to make it clearer though. Thanks @kwin and @manikmagar 👍

jonbullock:
Anyone else have any issue with this being merged in?

PR #543, opened 2018-10-10
temporary jruby dependency update

### LINKED ISSUE #542: Problems with asciidoctorj 1.5.7
With the latest upgrade to asciidoctorj 1.5.7 the jbake.org site does not bake properly.
See https://github.com/jbake-org/jbake.org/issues/106

It randomly fails to render either jbake.org/content/docs/2.4.0/configuration.adoc or jbake.org/content/docs/2.4.0/content-files.adoc.

(Other Versions of that file where rendered successful for `docs/2.2.1` for example)

It's not a problem of the file content. We use [String render(String content, Options options);](https://github.com/asciidoctor/asciidoctorj/blob/5079f5750aab780c37ca232e03a9835fcfaaf3c5/asciidoctorj-core/src/main/java/org/asciidoctor/Asciidoctor.java#L60) to render the contents.
See https://github.com/jbake-org/jbake/blob/b43537243a47c6517485194533ad387996732605/jbake-core/src/main/java/org/jbake/parser/AsciidoctorEngine.java#L158

The content in `context.getBody()` is not null at the point the Exception gets thrown.
And it contains valid asciidoc text.

ancho:
fixes #542

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/jbake-core/build.gradle
+++ b/jbake-core/build.gradle
@@ -14,6 +14,11 @@ dependencies {
     }
     compile "com.orientechnologies:orientdb-core:$orientDbVersion"
     compile "org.asciidoctor:asciidoctorj:$asciidoctorjVersion", optional
+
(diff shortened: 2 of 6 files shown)


ancho:
Ah...it's not compatible with jdk 7. So the only solution to fix 2.6.2 would be a downgrade to asciidoctorj 1.5.6 I guess -.-

jonbullock:
We could upgrade JRuby to v9.1.17.0 by declaring it as a dependency, it would work for JBake binary distribution but feels messy for the plugins and using JBake as an API.

Looks like the Asciidoctor Maven plugin has done this: https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/369/files

ancho:
Yes. We could do that as a temporary fix until the next release of asciidoctorj. I'll experiment with it and test against the gradle plugin.

ancho:
It works....let me add the smoketest against jbake.org...

ancho:
I squashed it to a version which tests the website and skips the test with jdk 7 for now.
It has serious problems with https using a compatible version of jgit to fetch the website repository.
See https://api.travis-ci.org/v3/job/440053613/log.txt for details.

PR #548, opened 2018-10-14
update to groovy 2.5.3 for JDK 11 support

ancho:
* update to groovy 2.5.3
* enable jdk 11 build at travis


jonbullock:
Looking forward to trying it out on JDK 11 👍

PR #564, opened 2018-11-09
Add current dir lib/* and lib/logging to classpath

maxandersen:
Why:

 *  when running jbake on cli there is no easy way to configure logging
    nor add additional jars.

This change addreses the need by:

 * extend current searh/replace of scripts to add `pwd`/lib
   and lib/logging

Note: I've tested the linux sh version on OSX, don't have a way
to test Windows nor pure linux right now but I have high confidence
that it will work :)


jonbullock:
Thanks Max 👍

PR #568, opened 2018-11-12
Catch all classloading issues

### LINKED ISSUE #567: TemplateEngines.tryLoadEngines should catch all throwables related to class loading issues
Currently in https://github.com/jbake-org/jbake/blob/aebf496e326151e91b9403522933c614d12e46a5/jbake-core/src/main/java/org/jbake/template/TemplateEngines.java#L80 only `Exception`s are being caught, but e.g. a `java.lang.NoClassDefFoundError` is not an `Exception` but only derived from a `Throwable`. Since all engines are IMHO optional even a `NoClassDefFoundError` should be caught and logged. For a concrete use cases where this is an issue look at https://github.com/jbake-org/jbake-maven-plugin/pull/17#issuecomment-428534051.

At the same time it would be reasonable to lower the log level to WARN in (https://github.com/jbake-org/jbake/blob/aebf496e326151e91b9403522933c614d12e46a5/jbake-core/src/main/java/org/jbake/template/TemplateEngines.java#L81) as it is perfectly fine if not all template engines are loaded in case they are not actually used.

kwin:
Lower loglevel to "WARN" as not all engines need to be loaded in all
circumstances.

This closes #567


jonbullock:
👍

PR #578, opened 2018-12-06
Fix certificates on OpenJDK 10 for Travis

jonbullock:


ancho:
That is to much pain to spend time with. I decided to drop 9 and 10 entirely in other PR's.
They reached EOL already. https://www.oracle.com/technetwork/java/java-se-support-roadmap.html
Let us safe energy and concentrate on LTS releases for CI.

jonbullock:
Ah OK, I'll scrap this PR then.

PR #580, opened 2019-01-06
Parse "jbake-" prefixed asciidoc headers like normal metadata

matthiasblaesing:
JBake explicitly supports adding complex metadata in the form of string
encoded JSON objects. This support was missing from metadata sourced
from native asciidoc headers. The support is only added for explicit
"jbake-" headers, to not interfere with normal asciidoc operation.

### DIFF
--- a/jbake-core/src/main/java/org/jbake/parser/AsciidoctorEngine.java
+++ b/jbake-core/src/main/java/org/jbake/parser/AsciidoctorEngine.java
@@ -98,7 +98,11 @@ public class AsciidoctorEngine extends MarkupEngine {
 
             if (hasJBakePrefix(key)) {
                 String pKey = key.substring(6);
-                documentModel.put(pKey, value);
+                if(value instanceof String) {
+                    processHeader(pKey, (String) value, documentModel);
+                } else {
+                    documentModel.put(pKey, value);
+                }
             }
             if (hasRevdate(key) && canCastToString(value)) {
 
--- a/jbake-core/src/main/java/org/jbake/parser/MarkupEngine.java
+++ b/jbake-core/src/main/java/org/jbake/parser/MarkupEngine.java
@@ -253,33 +253,35 @@ public abstract class MarkupEngine implements ParserEngine {
             if (hasHeaderSeparator(line)) {
                 break;
             }
-            processLine(line, context.getDocumentModel());
+            processHeaderLine(line, context.getDocumentModel());
         }
     }
 
-    private void processLine(String line, Map<String, Object> content) {
+    private void processHeaderLine(String line, Map<String, Object> content) {
         String[] parts = line.split("=", 2);
         if (!line.isEmpty() && parts.length == 2) {
+            processHeader(parts[0], parts[1], content);
+        }
+    }
 
+    void processHeader(String inputKey, String inputValue, Map<String, Object> content) {
+        String key = sanitizeKey(inputKey);
+        String value = sanitizeValue(inputValue);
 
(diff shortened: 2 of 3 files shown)


ancho:
Interesting. Thank you. I schedule some time for a review. Maybe end of the weekend.

jbake-core/src/main/java/org/jbake/parser/MarkupEngine.java
```diff
@@ -253,33 +253,35 @@ private void processDefaultHeader(ParserContext context) {
...
-                } catch (ParseException e) {
-                    LOGGER.error("unable to parse date {}", value);
-                }
-            } else if (key.equalsIgnoreCase(Crawler.Attributes.TAGS)) {
-                content.put(key, getTags(value));
-            } else if (isJson(value)) {
-                content.put(key, JSONValue.parse(value));
-            } else {
-                content.put(key, value);
+    void processHeader(String inputKey, String inputValue, Map<String, Object> content) {
```
ancho: I'm searching for a better name here. How about `processHeaderValue`

jbake-core/src/main/java/org/jbake/parser/AsciidoctorEngine.java
```diff
@@ -98,7 +98,11 @@ public void processHeader(final ParserContext context) {
 
             if (hasJBakePrefix(key)) {
                 String pKey = key.substring(6);
-                documentModel.put(pKey, value);
+                if(value instanceof String) {
+                    processHeader(pKey, (String) value, documentModel);
+                } else {
+                    documentModel.put(pKey, value);
```
ancho: We could always call `processHeader`. As the last case in that method adds the value for the given key to the documentModel Map. See line 284 in `MarkupEngine`.

jbake-core/src/main/java/org/jbake/parser/MarkupEngine.java
matthiasblaesing: I'm ok with changing the name, I used `processHeader` because this is key/value, an alternative would be `storeHeaderValue`, but I have no strong feeling into any direction.

jbake-core/src/main/java/org/jbake/parser/AsciidoctorEngine.java
matthiasblaesing: I decided to special case the "String" case, as I don't want to interfere with the asciidoc engine. As `NULL` could be generated, we'd still need a guard as in 
```java
if(value != null) {
    processHeader(pKey, value.toString(), documentModel);
}
```
I have zero experience with asciidocj, so can't predict if other types, than string are possible.

jbake-core/src/main/java/org/jbake/parser/MarkupEngine.java
ancho: Thats even a better name. That alignes with my attempt to build a document and template model abstraction. Which could be a new home for this method later. See #532

jbake-core/src/main/java/org/jbake/parser/AsciidoctorEngine.java
ancho: Yes you are right. asciidoctorj takes a Map<Sting, Object> for attributes. So it would be worth a try. On the other hand we could postpone it for a later cleanup session.

matthiasblaesing:
I pushed two updates, that try to address the comments. I kept them separate to make review easier, for integration I would squash them.

jonbullock:
Thanks for the contribution @matthiasblaesing 👍

Issue #391, opened 2017-07-05
[Feauture request] Improve baking speed


shalugin:
I want something like incremental site generation. I've enabled persistent content store (http://jbake.org/docs/2.5.1/#persistent_content_store). But generation can be faster.

For example we can do following things:
- Don't create `ContentStore` every time in `bake()` method.
- Skip copying assets, that already exists in destination folder.

jonbullock:
Thanks for raising this @shalugin 👍

jmcgarr:
I have also been looking into this feature, and I see a few ways to make this work. @jonbullock do you have an approach/design in mind already or do you mind if I fork and take a stab at it? I can see points of leverage being the `CustomFSChangeListener`, which currently does a bulk `Oven.bake()` rather than incrementally rerendering the content that changed. 

Also noticed that the `bake()` performance could improve by ensuring `assets` have already been copied and only copying those that have changed.

jmcgarr:
Sounds good @ancho! I agree with your three phases. I did more digging in yesterday and found just as many opportunities for improving performance in all three phases. I also like your proposal of making small pull requests to address the increased performance over time.

In taking a glance at tracking/caching asset management information, I was questioning the approach of OrientDB. In short, my suspicion is that spinning up the in-memory db and querying it incurs some performance penalty. My gut it to serialize/deserialize a Java object to a cache directory that can hold the state in memory. Thoughts on this approach?

I also noticed this issue was slated for jbake 3.0. Not sure if this changes the approach for accepting PR's?

ancho:
> In taking a glance at tracking/caching asset management information, I was questioning the approach of OrientDB. In short, my suspicion is that spinning up the in-memory db and querying it incurs some performance penalty. My gut it to serialize/deserialize a Java object to a cache directory that can hold the state in memory. Thoughts on this approach?

Hmm...I think that's because the Oven's `bake` method closes the database and shuts down the orient db engine at the end.

That's only a problem if running in serve mode. So it's basically deciding not to close and shutdown the database and let the caller of jbake handle that.

I have no interest in reinventing the wheel. If it's not fast enough if we don't shutdown and spin on the database we should consider to change the database.

Also...there is a newer versioin of orientdb available 3.0.8 which is not compatible with jdk 7.
So if we close #549 (pull-request in the making) we can see what we get out of it.
 
> I also noticed this issue was slated for jbake 3.0. Not sure if this changes the approach for accepting PR's?

[...]

PR #583, opened 2019-01-09
Incremental asset baking while serving

jmcgarr:
This is the first step to improving bake performance. The scope of this change is focused solely on copying newly changed asset files while serving content. This should work per asset file as it is changed. It should also copy asset files that exist in the content directory.

This pull request is a portion of the work contributing to #391.

### DIFF
--- a/jbake-core/src/test/java/org/jbake/app/AssetTest.java
+++ b/jbake-core/src/test/java/org/jbake/app/AssetTest.java
@@ -18,10 +18,12 @@ import java.net.URL;
 public class AssetTest {
 
     private DefaultJBakeConfiguration config;
+    private File fixtureDir;
 
     @Before
     public void setup() throws Exception {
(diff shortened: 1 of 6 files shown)


ancho:
Nice. I schedule some time for a review. Maybe end of the weekend. To much todo at present. Thank you.

jbake-core/src/test/java/org/jbake/app/AssetTest.java
```diff
@@ -152,6 +176,31 @@ public void testCopyAssetsFromContent(){
...
+        Assert.assertTrue(contentFile.exists());
+        Asset asset = new Asset(config);
+
+        Assert.assertTrue(asset.isAssetFile(cssAsset));
+        Assert.assertFalse(asset.isAssetFile(contentFile));
+    }
+
+    /*
+    @Test
+    public void testAssetTargetFolder() throws Exception {
```
ancho: Maybe remove that comment entirely. Or add an @Ignore with a description why it should be ignored for now.
What do you think?

jonbullock:
Thanks for the contribution @jmcgarr 👍

PR #598, opened 2019-02-19
add jcl-over-slf4j bridge

### LINKED ISSUE #597: make it possible to configure apache commons logging with logback
It seems that configuring lib/logging/logback.xml with `    <logger name="org.apache" level="DEBUG"/>
` does not have any affect unless the jcl-over.slf4j.jar is on the classpath.

https://www.slf4j.org/legacy.html

ancho:
makes it possible to configure `org.apache.commons` or
any 3rd party library using commons-logging with logback

closes #597

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -117,6 +117,7 @@ subprojects {
     dependencies {
         compile "org.slf4j:slf4j-api:$slf4jVersion"
         compile "org.slf4j:jul-to-slf4j:$slf4jVersion"
+        compile "org.slf4j:jcl-over-slf4j:$slf4jVersion"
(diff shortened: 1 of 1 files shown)


kwin:
Why has this milestone 2.7? It has been merged to master one year ago and just now 2.6.5 has been released from master...

ancho:
Yes. You are right. All closed issues in the v2.7.0 milestone are part of 2.6.5 now. We just missed to move them to 2.6.5 :facepalm: 

So 2.6.5 is more or less breaking the semantic versioning scheme. We should have created a realeas branch based on 2.6.4 and pick things. :shrug:

PR #614, opened 2019-10-28
Fixes #495: Added Freemarker escaping for urls.

### LINKED ISSUE #495: Set up URL escaping for Freemarker
Freemarker can escape URLs, but it needs the default encoding to be set.

    ${post.uri?url}

> To do URL encoding, the framework that encloses FreeMarker must specify the "output_encoding" setting or the "url_escaping_charset" setting, so ask the programmers to set them. Or, as a last chance, you can set the url_encoding_charset setting in the template, e.g. <#setting url_escaping_charset='ISO-8859-1'>, or give the charset explicitly to the built-in, e.g. foo?url('ISO-8859-1').
```
----
FTL stack trace ("~" means nesting-related):
	- Failed at: ${post.uri?url}  [in template "index.ftl" at line 10, column 34]
```

jaybob007:


jonbullock:
I realise it's been a while since this was raised, but thanks for the contribution.

PR #615, opened 2019-12-16
Feature/option improvements

toolforger:
Not 100% sure if this meets project standards, just let me know what needs to be done and what can be improved.


toolforger:
The failure looks like an old JVM trying to load a newer-format class file.

PR #624, opened 2020-03-26
Support Chinese characters in jbake.properties file by using UTF-8 encoding

georgecao:
JBake use Apache Configuration to load multi-properties files and also system properties. But right now PropertiesConfiguration reads the properties files in encoding ISO-8859-1, which is not support Chinese by default. Then when PropertiesConfiguration read property value in Chinese will cause the property missing due to not supported character encoding, it cannot decode key and value in lines containing Chinese characters.

In this change, I create `Utf8PropertiesConfiguration` class and override the `String getEncoding()` method and return UTF-8 to make it works.  The default `PropertiesConfiguration` load the file in the constructor, and there are no chances for us to specify a custom encoding before that.

And also, I excludes commons-logging from some deps.

### DIFF
--- a/jbake-core/build.gradle
+++ b/jbake-core/build.gradle
@@ -7,8 +7,12 @@ description = "The core library of JBake"
 
 dependencies {
     compile "commons-io:commons-io:$commonsIoVersion"
-    compile "commons-configuration:commons-configuration:$commonsConfigurationVersion"
-    compile "org.apache.commons:commons-vfs2:$commonsVfs2Version", optional
+    compile("commons-configuration:commons-configuration:$commonsConfigurationVersion") {
+        exclude group: "commons-logging", module: "commons-logging"
+    }
+    compile("org.apache.commons:commons-vfs2:$commonsVfs2Version") {
+        exclude group: "commons-logging", module: "commons-logging"
+    }
     compile "org.apache.commons:commons-lang3:$commonsLangVersion"
     compile("com.googlecode.json-simple:json-simple:$jsonSimpleVersion") {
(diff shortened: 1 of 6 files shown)


ancho:
Oh. Interesting. I was not aware and did not expect that. Thank you. Do you see a chance to make the encoding for the properties configurable? Like we do for `render.encoding`?

jonbullock:
What was the reasoning behind excluding commons-logging from the deps?

georgecao:
> What was the reasoning behind excluding commons-logging from the deps?

I see this PR #598, then commons-logging is not needed any more and should be exclude from the deps. But maybe a separate PR is better.

georgecao:
> Oh. Interesting. I was not aware and did not expect that. Thank you. Do you see a chance to make the encoding for the properties configurable, like we do for `render.encoding`?

OK

ancho:
Yes. It's a chicken/egg Problem. I would say, the default should be UTF-8. How about an additional cli option?

ancho:
Hmm....we are currently using commons 1.10 of apache commons. But there is also 2.x available. Maybe we should switch to the newer Version instead?
https://commons.apache.org/proper/commons-configuration/userguide/upgradeto2_0.html

Does not look that hard. Will play around with it.

georgecao:
> Yes. It's a chicken/egg Problem. I would say, the default should be UTF-8. How about an additional cli option?

Sounds good.

georgecao:
#640 solve this, I close this PR.

jonbullock:
Thanks for raising this in the first place @georgecao 👍

Issue #634, opened 2020-04-16
Include stack trace in log file when template engine can't be found


jonbullock:
Relates to #602 

Stack trace should be output in a log file somewhere even if it's not logged to the console.

pzygielo:
Please, consider changing level used in l.82 (to info or perhaps better - to debug) in
https://github.com/jbake-org/jbake/blob/4007fd2296ac03181fcf0fe32dd026d5c941bb0b/jbake-core/src/main/java/org/jbake/template/TemplateEngines.java#L74-L85

As current issue is continuation of #602, I'm not sure if this deserves separate one or could be covered here.

In my maven project following is reported
```
[INFO] --- jbake-maven-plugin:0.3.4:generate (build-site) @ xxx ---
[WARNING] Template engine not available: org.jbake.template.GroovyMarkupTemplateEngine
[WARNING] Template engine not available: org.jbake.template.PebbleTemplateEngine
[WARNING] Template engine not available: org.jbake.template.GroovyTemplateEngine
[WARNING] Template engine not available: org.jbake.template.ThymeleafTemplateEngine
[WARNING] Template engine not available: org.jbake.template.JadeTemplateEngine
[INFO] Baking has started...
```
as I have only freemarker added as dependency. (No other engines are needed.) Thus above warnings.

To get rid of them I could add the remaining engines, but I don't want to have unused dependencies to maintain.
If I have no required engine - I expect the error, and that's the case once I remove freemarker.

That's why I think it shall be either info, if it is so important, or rather debug only - to be used in investigation when actually needed engine can't be found. Not always.

PR #635, opened 2020-04-16
Removed stack trace from being output, and made log output more speci…

### LINKED ISSUE #602: Suppress "unable to load engine" warning for optional template engines
I'm trying to make the jbake-maven-plugin work with the latest version of JBake.
It's unfortunate that the JBake configuration API was changed incompatibly, but
I've mostly got it working.  However...

Most Maven projects don't include most template engines.  When JBake fails to
load an optional template engine, it produces a warning:

    [WARNING] unable to load engine

followed by a (huge) stack trace.  These warnings and stack traces obscure the fact
that JBake is working as expected.

JBake 2.6.0 did not produce any warning in this case.  These warnings should be removed,
or there should be a configuration option to suppress (or better yet, produce) these warnings.

The warning comes from [org.jbake.template.TemplateEngines](https://github.com/jbake-org/jbake/blob/master/jbake-core/src/main/java/org/jbake/template/TemplateEngines.java#L82).

jonbullock:
…fic by including reference to template engine.

Fixes #602.

### DIFF
--- a/jbake-core/src/main/java/org/jbake/template/TemplateEngines.java
+++ b/jbake-core/src/main/java/org/jbake/template/TemplateEngines.java
@@ -79,7 +79,7 @@ public class TemplateEngines {
             return ctor.newInstance(config, db);
         } catch (Throwable e) {
             // not all engines might be necessary, therefore only emit class loading issue with level warn
-            LOGGER.warn("unable to load engine", e);
+            LOGGER.warn("Template engine not available: {}", engineClassName);
             return null;
         }
     }


kwin:
I think completely hiding the stack prevents debugging issues. What about a earn without the stacktrace and emitting it still with level debug?

ancho:
Yes. You are right. We had a similar discussion leading to #634.
I think we should invest some time to cleanup logging in general and decide what is interisting from case to case. Lets disuss it further in #634

PR #639, opened 2020-04-27
feat: Allow the file used to ignore directories to be specified

rhwood:
This allows a single directory to be used by one JBake configuration and ignored by another JBake configuration by setting the `ignore.file` property in `jbake.properties`.

The default value for `ignore.file` is `.jbakeignore`.


manikmagar:
Thanks @rhwood for this PR. Could you please add some tests? For new config property, you could add in `ConfigUtilTest`.

PR #640, opened 2020-04-28
update to apache configurations2. load properties files with default utf-8 encoding

ancho:
This is an update to apache commons configuration 2.7 which is up to date and much more configurable than 1.10.

You can now configure the encoding that should be used loading the properties file in a fluent way.

This PR makes #624 obsolete. But without the original PR we wouldn't even know that 1.10 had an encoding issue.

The only thing that is missing:

- [x] add cli option (`--prop-encoding`) to configure property encoding that should be used


ancho:
What's wrong with windows builds and utf-8?

ancho:
@jonbullock , @manikmagar 
Ok. I give up. I'm to stupid fixing the appveyor windows builds. Any ideas?

ancho:
All right. Setting jvm parameter file.encoding to UTF-8 for windows solves the issue.
`./gradlew -Dfile.encoding=UTF-8 -i check`
I will cleanup/sqash the mess.

jbake-core/src/main/java/org/jbake/app/configuration/DefaultJBakeConfiguration.java
```diff
@@ -89,15 +88,15 @@ private String getAsString(String key, String defaultValue) {
...
         Configuration subConfig = compositeConfiguration.subset(JBakeProperty.ASCIIDOCTOR_OPTION);
-        Object value = subConfig.getProperty(optionKey);
 
-        if (value == null) {
+        if (subConfig.containsKey(optionKey)) {
+            return subConfig.getList(String.class, optionKey);
+        } else {
             logger.warn("Cannot find asciidoctor option '{}.{}'", JBakeProperty.ASCIIDOCTOR_OPTION, optionKey);
-            return "";
+            return new ArrayList<>();
```
manikmagar, line 98: Should this be `Collections.emptyList()` to return immutable list?

jonbullock:
👍

PR #646, opened 2020-05-31
Wrap usage at 80 characters instead of 100

### LINKED ISSUE #645: `jbake -h` output is too wide
The usage statement is wrapped at 100 characters even though most terminals default to a width of 80 characters. Is there a reason this is set to 100 and not 80?

Screen Shot 2020-05-31 at 17 03 10

rhwood:
Fixes #645


ancho:
I will add your recommendations to #572 
We consider to switch to picocli.

PR #647, opened 2020-05-31
feat: allow config file to be specified on command line

rhwood:
Adds the `-c VAL` or `--config VAL` command line arguments to allow VAL to be used as a configuration file instead of `jbake.properties`.

Combined with #639, this allows a single source to be used for multiple purposes (for example, the help content of an app could be packaged for a site using one set of templates and for the app using a different set of templates).


jonbullock:
Thanks for submitting this PR @rhwood 👍

jonbullock:
Merged in via #700

PR #648, opened 2020-06-12
Don't assume document types are valid identifiers in OrientDB

matthiasblaesing:
The ContentStore implementation assumes, that the document types are
valid identifiers, that can be used without quoting them. This limits
document types to "simplified identifiers". This changeset quotes the
identifiers and thus allows for example using a hyphen in the document
type.

This problem was observed when trying to update an older setup, that
was using types like "page-front", "page-noaside", "platform-tutorial".

### DIFF
--- a/jbake-core/src/main/java/org/jbake/app/ContentStore.java
+++ b/jbake-core/src/main/java/org/jbake/app/ContentStore.java
@@ -55,6 +55,7 @@ import java.util.Set;
 public class ContentStore {
 
     private static final String STATEMENT_GET_PUBLISHED_POST_BY_TYPE_AND_TAG = "select * from %s where status='published' and ? in tags order by date desc";
+    private static final String STATEMENT_GET_POST_BY_TYPE_AND_URI = "select * from %s where sourceuri=?";
     private static final String STATEMENT_GET_DOCUMENT_STATUS_BY_DOCTYPE_AND_URI = "select sha1,rendered from %s where sourceuri=?";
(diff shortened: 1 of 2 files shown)


matthiasblaesing:
With this fix in I was able to rebuild the project again (updated versions for gradle, jbake gradle plugin and jbake itself) and at least from a first look the result looked identical.

matthiasblaesing:
Anything I can help to get this into a release? This is a hard update block for me, that pin the build to JDK8.

ottlinger:
@jonbullock @ancho can you help out :) ?! Thanks.

jonbullock:
Thanks, I'll make some time to review over the weekend.

jonbullock:
Thanks @matthiasblaesing for submitting this improvement, the plan is to include this in 2.7.0

matthiasblaesing:
@jonbullock thank you for checking. I'm not sure though, that I agree with the assessment, that this is an enhancement. In the past it was possible to use document types with hyphens in the name. That was broken somewhere between 2.5.0 and 2.6.5, so I would call it a regression.

Can you estimate a timeframe when 2.7.0 might be released?

jonbullock:
Ah my mistake, I misread the issue. I've created a v2.6.7 milestone, as a bug fix release, and assigned this to it.

matthiasblaesing:
@jonbullock can you estimate, when a bugfixed version could be available?

jonbullock:
@matthiasblaesing apologies for the delay in getting back to you on this, however we're aiming to get v2.6.6 released very soon.

PR #649, opened 2020-06-15
Support xref aka cross reference link fix in adoc

georgecao:
I use JBake version 2.6.5 and adoc to write some blog. JBake has two configuration properties to fix relative image src. And they are `img.path.update` and `img.path.prepend.host`. These two properties only affect `img` tag and it's `src` attribute.

But in adoc, we can add a reference in one doc to another by for example` xref:a-gentle-introduction-to-multithreading.adoc[introduction] `, like image , this will generate a `<a href='a-gentle-introduction-to-multithreading.html''>...</>` tag with relative path. This need fix too.

I introduce three new properties, `relative.path.update`, `relative.path.prepend.host` and `relative.tag.attribute`. `relative.path.update` and `relative.path.prepend.host` are for configuring all relative paths updates, not just `img` tag. And the third property `relative.tag.attribute` defines which tags need fix, for example `relative.tag.attribute=img=src,a=href`.

For backward compatible, both the new and old properties should work.

With a false default value, we should use the properties this way
`boolean pathUpdate = config.getImgPathUpdate() || config.getRelativePathUpdate()`
With a true default value, we should use the properties this way:
`boolean prependSiteHost = configuration.getImgPathPrependHost() && configuration.getRelativePathPrependHost();`

### DIFF
--- a/jbake-core/src/main/java/org/jbake/app/Crawler.java
+++ b/jbake-core/src/main/java/org/jbake/app/Crawler.java
@@ -213,9 +213,9 @@ public class Crawler {
                     fileContents.put(Attributes.NO_EXTENSION_URI, uri.replace("/index.html", "/"));
                 }
 
-                if (config.getImgPathUpdate()) {
-                    // Prevent image source url's from breaking
-                    HtmlUtil.fixImageSourceUrls(fileContents, config);
+                if (config.getImgPathUpdate() || config.getRelativePathUpdate()) {
+                    // Prevent image or other tag's source url's from breaking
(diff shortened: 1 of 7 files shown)


jonbullock:
Thanks for taking the time to raise this PR. 

This area of functionality will be reviewed, hopefully for the 2.7.0 release, so it may be a while before this gets properly reviewed.

georgecao:
> Thanks for taking the time to raise this PR.
> 
> This area of functionality will be reviewed, hopefully for the 2.7.0 release, so it may be a while before this gets properly reviewed.

Good to know this.

PR #671, opened 2021-03-21
publish to sonatype nexus instead of bintray

ancho:
This is an attempt to replace the bintray publishing with a plugin to directly publish the artifacts to sonytype nexus.
It uses the [gradle nexus publish plugin](https://github.com/gradle-nexus/publish-plugin).

There are three things that need to be handled.

- [x] publish jbake-core lib to sonatype nexus
- [x] publish the binary distribution somewhere
- [x] Fix location for the [sdkman publishing task](https://github.com/jbake-org/jbake/blob/master/gradle/sdkman.gradle#L12)

The question is where to put the binary distribution?

We could use create github realeases and add the distribution there or copy it to the webpage of jbake.
The latter requires some git kungfu, i guess. As the jbake.org page lives at gitlab pages.

Currently i am more into using github releases but I am open for other ideas too.


ancho:
First experiments successful with the github releases plugin.

Screenshot-area-2021-03-21T15:15:10+01:00

jonbullock (approved):

jonbullock:
Makes sense to use GitHub releases for the binary distribution so it's available alongside the source code, so lets go with it.

manikmagar:
Using Github releases makes more sense for github source. It is easy to find binaries. I have been using Github Actions to release to github as well as to publish packages Sdkman, homebrew, scoop, container etc  - https://github.com/manikmagar/mulefd/blob/master/.github/workflows/publish-packages.yml. I am now keeping an eye on this https://github.com/jreleaser/jreleaser for possibly simplifying that workflow.

jonbullock:
Yeah jreleaser is on my list of things to check out.

aalmiray:
I can help getting JReleaser configured in this build 😏

PR #672, opened 2021-04-04
produce reproducible archives (distribution zip and jar files)

ancho:
Using the timestamp of the current repository HEAD and make sure the generated jar and zipfiles evaluate to the same checksums after rebuilding.


jonbullock:
Makes so much more sense to use the date/time of the repo compared to when the build was executed  👍

PR #681, opened 2021-04-26
Added Dockerfile and updated README

jonbullock:
Resolves jbake-org/jbake-docker-image#2 jbake-org/jbake-docker-image#3

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8-alpine as builder
+FROM adoptopenjdk/openjdk11:alpine as builder
 
 LABEL maintainer="https://jbake.org/community/team.html"
 
@@ -14,7 +14,7 @@ RUN set -o errexit -o nounset \
     && cp -r jbake-dist/build/install/jbake/* $JBAKE_HOME \
     && rm -r ~/.gradle /usr/src/jbake
 
-FROM openjdk:8-jre-alpine
+FROM adoptopenjdk/openjdk11:alpine-jre
 
 ENV JBAKE_USER=jbake
 ENV JBAKE_HOME=/opt/jbake
@@ -31,6 +31,6 @@ WORKDIR /mnt/site
 VOLUME ["/mnt/site"]
 
 ENTRYPOINT ["jbake"]
-CMD ["-b","-s"]
+CMD ["-b"]
 
 EXPOSE 8820
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,36 @@
+FROM openjdk:8-alpine as builder
+
+LABEL maintainer="https://jbake.org/community/team.html"
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+USER ${JBAKE_USER}
+
+COPY --from=builder /opt/jbake /opt/jbake
+
+WORKDIR /mnt/site
+
+VOLUME ["/mnt/site"]
+
+ENTRYPOINT ["jbake"]
+CMD ["-b","-s"]
```
manikmagar, line 34: Should we put `-s` in the command? Most of the use case could be run this as a part of build pipeline. 
To just bake the site without serving `-b` argument will need to be provided to override default CMD - 

```
docker run --rm -u jbake -v "$PWD":/mnt/site jbake/jbake:latest -b
```

Other option is to default it to "-b" but then users who will use it to server locally, will require to override CMD with `-b -s`.

Eitherway, I think we should document this in README.

ancho: Yes. That's a good idea.

Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM openjdk:8-alpine as builder
```
ancho, line 1: We should switch to a more docker friendly version of the jdk/jre images.
Smaller footprint. Better jvm memory consumption.
I started playing around with the latest version from adoptopenjdk which works very well.
See https://github.com/ancho/jbake/blob/feature/docker-multistage-build/Dockerfile

ancho (changes requested):

Dockerfile
jonbullock: Yeah good point, I'll default to just `-b` and update the README.

Dockerfile
jonbullock: Yeah happy to go with adoptopenjdk, I'll update now.

jonbullock: Had to revert to adoptopenjdk/openjdk11 as Gradle 6.8.3 didn't like openjdk16, but noticed your branch had an update to Gradle 7 so we can upgrade to openjdk16 at the same time.

README.asciidoc
```diff
@@ -54,6 +54,41 @@ Talk to other users of JBake on the forum:
 
 * http://groups.google.com/group/jbake-user[Forum]
 
+== Docker Image
+
+The image uses the official openjdk:8-alpine image for building a distribution of JBake and openjdk:8-jre-alpine for runtime.
```
manikmagar, line 59: This doesn't look correct anymore. New change is using JDK11 to build and JRE11 to run.

manikmagar (changes requested):

README.asciidoc
jonbullock: Ah yeah thanks for catching that.

manikmagar (approved):
LGTM

Issue #664, opened 2021-02-28
org.jbake.template.PebbleTemplateEngine


rdmueller:
I always get a warning `Template engine not available: org.jbake.template.PebbleTemplateEngine`
Is there a way to supress this?

I use the jbake plugin for gradle (5.3.0) with jbake 2.6.5

ancho:
Strange. I didn't notice so far.
Does the warning appear with jbake cli as well? 
Could be some strange classpath situation with gradle. The class is there https://github.com/jbake-org/jbake/blob/v2.6.5/jbake-core/src/main/java/org/jbake/template/PebbleTemplateEngine.java

What gradle version are you using in your project?

ancho:
Jup. The pebble dependency is not given by the gradle plugin and the PebbleTemplateEngine fails during initialization.

Add the following to your build.gradle file:

```
dependencies {
    jbake "io.pebbletemplates:pebble:3.1.2"
}
```

ancho:
I wonder if the plugin should solve the missing dependency, or if this situation should be handled somehow by the library.
Like enabling/disabling certain engines via configuration to make the optional dependencies really optional ;)
It should definitly give a more meaningful error message if it fails though.

rdmueller:
thanx. the fix given works for me and I am fine with this workaround. you can close this issue now if you like.

rdmueller:
btw: greate response time! :-D

jonbullock:
Linking related issues: #602 #634

jonbullock:
Closed via #682

PR #682, opened 2021-04-29
Use lower log level to report missing template engine

pzygielo:
For the reasons presented in https://github.com/jbake-org/jbake/issues/634#issuecomment-827036102


jonbullock:
Thanks for taking the time to contribute this change.

jonbullock:
This also addresses #664

PR #702, opened 2021-05-23
Change config interface to simplify build plugins

jonbullock:
Added extra method so build plugins don't need to depend on Apache Commons Configuration library.


jonbullock:
Thanks @ancho for catching the Exception type thrown!

PR #720, opened 2021-08-28
Fix relative url 

georgecao:
Fix path url that contains ../../../


jonbullock:
@georgecao Was this to do the same for links as we already had for img src (https://jbake.org/docs/2.6.7/#images_path)?

PR #728, opened 2021-10-03
WIP: Bake faster

ancho:
This is a Proof of Concept to speed up the baking process using parallel execution.

It speeds up the following three core activities:

* crawling
* rendering
* asset management (copy stuff around)

This is just a quick sketch to start a discussion on possible solutions on how to get more performance out of the baking process.

I kept it as simple as possible to just get it working. 
But the code needs to be cleaned up/improved a little bit. And there are missing tests regarding possible side effects.
I think it would be better to walk the file tree instead of using a parallel stream with recursive calls for example.

But first I would be happy to know what others think about the current approach.

### DIFF
--- a/jbake-core/src/main/java/org/jbake/app/ContentStore.java
+++ b/jbake-core/src/main/java/org/jbake/app/ContentStore.java
@@ -302,9 +302,11 @@ public class ContentStore {
         return DocumentList.wrap(results);
     }
 
-    private void executeCommand(String query, Object... args) {
+    private synchronized void executeCommand(String query, Object... args) {
         activateOnCurrentThread();
+        db.getTransaction().begin();
         db.command(query, args);
+        db.getTransaction().commit();
     }
 
     public Set<String> getTags() {
@@ -407,10 +409,12 @@ public class ContentStore {
         return db.isActiveOnCurrentThread();
     }
 
-    public void addDocument(DocumentModel document) {
+    public synchronized void addDocument(DocumentModel document) {
+        db.getTransaction().begin();
         ODocument doc = new ODocument(Schema.DOCUMENTS);
         doc.fromMap(document);
         doc.save();
+        db.getTransaction().commit();
     }
 
     protected abstract class Schema {
--- a/jbake-core/src/main/java/org/jbake/app/ContentStore.java
+++ b/jbake-core/src/main/java/org/jbake/app/ContentStore.java
@@ -169,8 +169,6 @@ public class ContentStore {
 
     public void drop() {
         activateOnCurrentThread();
-//        db.drop();
-
         orient.drop(name);
     }
 
@@ -290,13 +288,13 @@ public class ContentStore {
         executeCommand(STATEMENT_INSERT_TEMPLATES_SIGNATURE, currentTemplatesSignature);
     }
 
-    private DocumentList<DocumentModel> query(String sql) {
+    private synchronized DocumentList<DocumentModel> query(String sql) {
         activateOnCurrentThread();
         OResultSet results = db.query(sql);
         return DocumentList.wrap(results);
     }
 
-    private DocumentList<DocumentModel> query(String sql, Object... args) {
+    private synchronized DocumentList<DocumentModel> query(String sql, Object... args) {
         activateOnCurrentThread();
         OResultSet results = db.command(sql, args);
         return DocumentList.wrap(results);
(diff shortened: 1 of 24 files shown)


jbake-core/src/main/java/org/jbake/app/ContentStore.java
```diff
@@ -290,21 +283,23 @@ private void insertTemplatesSignature(String currentTemplatesSignature) {
...
         OResultSet results = db.command(sql, args);
         return DocumentList.wrap(results);
     }
 
-    private void executeCommand(String query, Object... args) {
+    private synchronized void executeCommand(String query, Object... args) {
         activateOnCurrentThread();
+        db.getTransaction().begin();
         db.command(query, args);
+        db.getTransaction().commit();
```
bmarwell, line 302: Why use synchronize here? Instead, you could catch `OConcurrentModificationException` and retry.

See: https://orientdb.org/docs/3.0.x/general/Concurrency.html

jbake-core/src/main/java/org/jbake/app/DocumentRenderAgent.java
```diff
@@ -0,0 +1,66 @@
...
+            outputFilename = outputFilename + config.getDraftSuffix();
+        }
+
+        File outputFile = new File(outputFilename + outputExtension);
+        TemplateModel model = new TemplateModel();
+        model.setContent(document);
+        model.setRenderer(renderingEngine);
+
+        try {
+            try (Writer out = createWriter(outputFile)) {
```
bmarwell, line 52: Is this double-nesting needed? I do not see a reason.

jbake-core/src/main/java/org/jbake/app/Oven.java
```diff
@@ -154,38 +154,39 @@ public void bake() {
...
+            LOGGER.info("Rendering content...");
             renderContent();
 
             // copy assets
+            LOGGER.info("Copy assets...");
             asset.copy();
             asset.copyAssetsFromContent(config.getContentFolder());
 
+            utensils.getRenderer().shutdown();
             errors.addAll(asset.getErrors());
```
bmarwell, line 178: shutdown() does not wait for the ES to run all threads until finished (ie non-blocking operation). You need for all threads to finish, or you will not get all asset errors.

jbake-core/src/main/java/org/jbake/app/Oven.java
```diff
@@ -154,38 +154,39 @@ public void bake() {
...
             LOGGER.info("Baking finished!");
-            long end = new Date().getTime();
-            LOGGER.info("Baked {} items in {}ms", renderedCount, end - start);
+            long bakeEnd = new Date().getTime();
+            LOGGER.info("Baked {} items in {}ms", utensils.getRenderer().getRenderCount(), bakeEnd - bakeStart);
             if (!errors.isEmpty()) {
                 LOGGER.error("Failed to bake {} item(s)!", errors.size());
             }
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
bmarwell, line 187: * Re-Interrupt
* use logger

jbake-core/src/main/java/org/jbake/app/Oven.java
```diff
@@ -216,7 +217,7 @@ private void resetDocumentTypesAndExtractors() {
     /**
      * Load {@link RenderingTool} instances and delegate rendering of documents to them
      */
-    private void renderContent() {
+    private void renderContent() throws InterruptedException {
```
bmarwell, line 220: The method never checks if it got interrupted. The for loop would be a sane choice.

jbake-core/src/main/java/org/jbake/app/Renderer.java
```diff
@@ -344,7 +293,40 @@ private DocumentModel buildSimpleModel(String type) {
...
+    }
+
+    public List<Throwable> getErrors() {
+        return errors;
+    }
+
+    public void shutdown() throws InterruptedException {
+        this.excutor.shutdown();
+        this.excutor.awaitTermination(10, TimeUnit.MINUTES);
+    }
```
bmarwell, line 307: Ah, you made it synchronous. Why wait here? Better name it "shutdownAndWait" then, because usually the shutdown method does not wait.

bmarwell (commented):
Some quick findings. I wonder if it would be better to use a global ExSvc instead of shutting new ones down all the time (see comments about waiting).

PR #731, opened 2021-11-08
Improve GRADLE build Performance

shisheng-1:
[File system watching](https://blog.gradle.org/introducing-file-system-watching). Since Gradle 6.5, File system watching was introduced which can help to avoid unnecessary I/O. This feature is the default since 7.0. For an older version, we can enable this feature by setting `org.gradle.vfs.watch=true`.

[Configuration on demand](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:configuration_on_demand). Configuration on demand tells Gradle to configure modules that only are relevant to the requested tasks instead of configuring all of them. We can enable this feature by setting `org.gradle.configureondemand=true`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -45,3 +45,5 @@ grgitVersion                = 4.1.0
 
 org.gradle.caching=true
 org.gradle.parallel=true
+org.gradle.configureondemand = true
+org.gradle.vfs.watch = true


jonbullock:
Thanks for the contribution :+1:

PR #737, opened 2021-12-24
Configure JReleaser

aalmiray:
This PR configures JReleaser for creating Git releases and publishing distributions to Sdkman. 
It also adds GitHub workflows for:

 - building on PRs
 - publishing early-access builds

A workflow for publishing a final release is pending.

The following enviroment variables are required

- JRELEASER_GITHUB_TOKEN
- JRELEASER_GPG_PASSPHRASE
- JRELEASER_GPG_PUBLIC_KEY
- JRELEASER_GPG_SECRET_KEY

Values for these variables may be stored at a local file in `.properties`, `.yml`, `.toml`, or `.json` format. Refer to https://jreleaser.org/guide/latest/configuration/environment.html Set these values if you intended to test releases locally. You may also set these values as repository secrets.

Refer to https://jreleaser.org/guide/latest/configuration/signing.html for how to configured GPG signing


aalmiray:
converting to draft as it requires JReleaser v0.10.0, to be released in the coming days.

jonbullock:
I'll mark this for inclusion in the 2.7.0 release.

aalmiray:
@jonbullock ready for merge

.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Java
+        uses: actions/setup-java@v2
+        with:
+          java-version: 11
+          distribution: 'zulu'
+
+      - uses: actions/cache@v2
```
kwin, line 24: This is supported by setup-java as well: https://github.com/actions/setup-java#caching-packages-dependencies

aalmiray: Yes, it is. Don't count on that feature to be permanent, as you may have noticed in this issue https://github.com/actions/setup-java/issues/40#issuecomment-999730509

kwin: I am reporter of https://github.com/actions/setup-java/issues/40 and this is only about configuring Maven but not about caching at all (particularly not about caching Gradle artifacts)

aalmiray: I'm well aware of that which is why I typed "as you may have noticed". IMHO `setup-java` should be concerned with just that: setting up a suitable JDK/JRE/JR. Adding caching for Maven or any other Java build tool is beyond its scope.

kwin: This may happen, but never in a 2.x release of setup-java (is it would be backwards-incompatible). As long as you use setup-java@v2 I would recommend getting rid of additional caching action steps and leverage what is shipped with setup-java@v2 as having additional steps
# make syour yml larger
# doesn't provide additional functionality
# defers builds as every step is additional overhead

jonbullock:
I'll manually merge this in now #740 has been merged in before it.

PR #740, opened 2021-12-28
Add repo jbake-maven-plugin into repo jbake + upgrade to gradle 7

ge0ffrey:
- [X] Create subdirectory jbake-maven-plugin with the content of the old repo
- [X] Don't lose git history or attribution
- [X] Build maven plugin as part of the build
- [X] Upgrade to Gradle 7.x to the previous to work (Thanks Ancho!)
- [X] Documentation and release notes - PR for that: https://github.com/jbake-org/jbake.org/pull/150
- [X] Example using the plugin (proof that it works well too) - PR for that: https://github.com/jbake-org/jbake-example-project-freemarker/pull/14

Tackles these issues:
1) A JBake release should automatically release the maven and gradle plugin too. More efficient releases.
2) The version number of the maven and gradle plugins should match the JBake versions.
Keep it simple for users: If a new JBake version comes out, I just update my jbake-maven-plugin version, done. **No need to figure out which maven/gradle plugin version aligns to which jbake version** (for every jbake upgrade too).
3) The Maven and Gradle versions should use the same asciidoc, markdown, freemarker, etc versions as jbake-core. They should include them all out of the box, so a user doesn't have to explicitly add them in their build files. They shouldn't have to figure out the right asciidoc and freemarker version to use, nor go through all that again every time when upgrading to a new jbake version. The should automatically use the same asciidoc/markdown version as jbake-core unit tests did.
4) All of this begins with moving the maven and gradle repositories into the gradle repository without loosing git history.


lefou (commented):
What is the motivation of this merge?

Some thoughts, why I question this move:
* The jbake-maven-plugin has a different lifecycle than JBake artifacts
* not every JBake release requires a new jbake-maven-plugin release
* newer releases of the jbake-maven-plugin might be necessary to keep up with changes in Maven, without requiring a JBake release
* Maven plugins are best built with Maven, to keep low level entrance for Maven folks, whereas JBake is built with Gradle

ge0ffrey:
> What is the motivation of this merge?

Sorry for not referencing the motivation for this PR here. There's a long discussion the JBake chat. I have adjusted the description of this PR with a TLDR.

> Some thoughts, why I question this move:
> 
>     * The jbake-maven-plugin has a different lifecycle than JBake artifacts

This PR unifies the lifecycle of jbake-core, jbake-maven-plugin (also in groupId org.jbake) and (coming soon) jbake-gradle-plugin (which will refactor to groupId org.jbake).
They will be released all together, using the same version.

When a user gets a notification that JBake is released, he/she can simply flip their jbakeVersion=2.7.0 to 2.8.0 and they're done. No need to maintain jbakeCoreVersion, jbakeMavenVersion, ... and align those for every release.

> 
>     * not every JBake release requires a new jbake-maven-plugin release

Actually it does:
- All unit/integration tests in jbake-maven-plugin need to run again with the new jbake-core version to ensure that the new jbake-core version doesn't break the maven plugin.
- Users that want to consume the new jbake-core version should not have to overwrite the jbake-maven-plugin's dependencies to get the new version. That is a very bad practice. They also don't need the complexity of managing 2 different jbake versions (jbakeCoreVersion and jbakeMavenVersion).

> 
>     * newer releases of the jbake-maven-plugin might be necessary to keep up with changes in Maven, without requiring a JBake release

No users care. They don't want to juggle multiple jbake-foo and jbake-bar versions. One Jbake version to rule them all.
A new feature in any jbake component (including the jbake-maven-plugin) warrents a new minor jbake version, that applies to all jbake components and is **released and tested as a whole**.

> 
>     * Maven plugins are best built with Maven, to keep low level entrance for Maven folks, whereas JBake is built with Gradle

Agreed. On the other hand, contributors that touch the jbake-core should be building both the maven and gradle plugins - as wells as CI - to ensure they don't break those.
Given this trade-off, building the maven plugin with gradle is the lesser evil.

lefou:
This sounds to me like some unfortunate design choices made in the plugin, resulting in a tightly coupling to concrete JBake versions. Working around this by making the tight coupling more obvious (by merging the repos) is only one of the possible solutions. The outcome (building a Maven plugin with Gradle) is a rather complication beast and will be hard to maintain.

But it looks like decisions are already made. Also, as I'm not doing any of this work, I'll mute myself. Sorry for the interruption.

ge0ffrey:
> This sounds to me like some unfortunate design choices made in the plugin, resulting in a tightly coupling to concrete JBake versions. Working around this by making the tight coupling more obvious (by merging the repos) is only one of the possible solutions. The outcome (building a Maven plugin with Gradle) is a rather complication beast and will be hard to maintain.

Good point, that would be better, at a development cost. Currently it has semi loose coupling, but for full loose couping, jbake-core would need a Backward Compatible api, preferably in a jbake-api jar, fully javadocced, etc.
I would argue YAGTNI - we ain't going to need it: loose coupling between jbake-maven-plugin and jbake-core would require a matrix of unit tests and would probably still not avoid the **compatibility matrix**.

> But it looks like decisions are already made. 

Not really. This is Jon's call, but I did manage to convince him that doing this POC is worthwhile. I hope I can convince you too.
As a contributor, I just care very strongly about this, here's why:
  https://github.com/kiegroup/optaplanner-website/blob/main/optaplanner-website-root/pom.xml#L19
This issue makes every jbake upgrade a non-trivial amount of work - to align all versions - while it should be trivial.

lefou:
> Good point, that would be better, at a development cost. Currently it has semi loose coupling, but for full loose couping, jbake-core would need a Backward Compatible api, preferably in a jbake-api jar, fully javadocced, etc.
> I would argue YAGTNI - we ain't going to need it: loose coupling between jbake-maven-plugin and jbake-core would require a matrix of unit tests and would probably still not avoid the **compatibility matrix**.

Actually, JBake has a rather stable CLI. And even supporting multiple CLI's depending on the version is rather simple. In the past I made sure that JBake CLI can also be used as library, you just need to avoid the "main" method, which uses `System.exit`. To avoid JVM startup costs, one can load all required JBake libraries in an isolated classloader for repeated use. This is of course just one approach, but it results in a rather simple plugins that require only small maintenance effforts.It is also very similar to using pure JBake without any integration, so it's optimal for users in understanding the setup in case anything goes wrong or to reproduce and report issues. That's why I had choosen this approach in my JBake integrations. But I need to admin, I haven't used JBake for a while now.

jonbullock:
Thanks @lefou for taking the time to share your thoughts on this, it is very much appreciated. 👍 

There are negatives, as you quite rightly point out, but I think the potential positives will outweigh the negatives overall. Especially for users and that is the goal for this.

jbake-core/build.gradle
```diff
@@ -1,32 +1,38 @@
...
 description = "The core library of JBake"
 
 dependencies {
-    implementation "commons-io:commons-io:$commonsIoVersion"
-    implementation "org.apache.commons:commons-configuration2:$commonsConfigurationVersion"
+    api "commons-io:commons-io:$commonsIoVersion"
+    api "org.apache.commons:commons-configuration2:$commonsConfigurationVersion"
     implementation "commons-beanutils:commons-beanutils:$commonsBeanutilsVersion"
-    implementation "org.apache.commons:commons-vfs2:$commonsVfs2Version", optional
+    implementation "org.apache.commons:commons-vfs2:$commonsVfs2Version"
```
jonbullock, line 19: Was this optional removed intentionally?

ge0ffrey: No, let me restore that

ge0ffrey: Fixed.

jonbullock: :+1:

jonbullock:
Conducted some local testing and the plugin builds and executes as expected in a sample Maven project. I compared the generated artefact to a previous version and noticed a few things. The pom.xml for the plugin is missing metadata about the plugin (name, license, SCM etc.) and a HelpMojo class is missing (guessing this was automatically generated).

Couldn't remember if the metadata is required for Maven Central so did a check and found this: https://central.sonatype.org/publish/requirements/ which suggests it is.

ancho:
Have a look at the latest commit here https://github.com/ancho/jbake/tree/merge-repo-jbake-maven-plugin
I had to switch the maven development plugin to make it work.

jonbullock:
Thanks @ancho that's solved it, with that commit I get the metadata and the HelpMojo class.

ge0ffrey:
Thanks  @ancho! Great work!
@jonbullock I fast forwarded this PR to include ancho's last commit too.

jonbullock:
Thanks so much @ge0ffrey and @ancho

PR #757, opened 2022-03-20
Resolves #747 - OTrackedMap returned for contents of data file

### LINKED ISSUE #747: data.get('file.yaml') will return `class com.orientechnologies.orient.core.db.record.OTrackedMap`
for example, given this file:
https://github.com/apache/shiro-site/blob/6b06cb888e04febf8216d82c1043fb27eea45044/data/authors.yaml

## actual behaviour

When importing using
`<#assign authors = data.get('authors.yaml').authors />`

The returned type will be `com.orientechnologies.orient.core.db.record.OTrackedMap`

## expected behaviour

A hash is returned without methods like `getClass()`. Just the author names as keys and the data as values.

## version

```xml
          <dependency>
            <groupId>org.jbake</groupId>
            <artifactId>jbake-core</artifactId>
            <version>2.7.0-rc.4</version>
          </dependency>
```

jonbullock:
Fixes #747 - Force the OrientDB type rather than let OrientDB decide.

### DIFF
--- a/jbake-core/src/main/java/org/jbake/app/ContentStore.java
+++ b/jbake-core/src/main/java/org/jbake/app/ContentStore.java
@@ -32,10 +32,8 @@ import com.orientechnologies.orient.core.db.OrientDBConfig;
 import com.orientechnologies.orient.core.metadata.schema.OClass;
 import com.orientechnologies.orient.core.metadata.schema.OSchema;
 import com.orientechnologies.orient.core.metadata.schema.OType;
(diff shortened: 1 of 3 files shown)


bmarwell (commented):
Haven’t tried it locally (I do not see a test case either). Do you need one?

jonbullock:
Yeah just in case of a regression, I'll add a check for this scenario.

bmarwell (approved):

PR #758, opened 2022-03-24
As adoptopenjdk is deprecated the dependencies should be adopted to its successor adoptium/temurin

areyouready:


jonbullock:
Thanks for the contribution :+1:

PR #764, opened 2022-08-28
[#763] Allow overriding jbake.properties through maven plugin

### LINKED ISSUE #763: Allow overriding properties for Maven plugin through properties configuration
jbake.properties should be overridable through the configuration section of the maven plugin. Also see https://github.com/jbake-org/jbake-maven-plugin/issues/44

beikov:
Fixes #763


johncarl81:
This would be a great addition to the plugin.  In the mean time, is there a recommended way to use maven properties as jbake template variables?

